PR #118, opened 2014-09-29
WIP: ionic start with support for Creator apps

mlynch:


jbavari:
It's looking like this was already included. Closing.

PR #132, opened 2014-10-30
adds proxy support to the serve command

blackjid:
This PR enables the serve command to proxy your api server though the ionic server. This way you don't have the cross origin problems while you develop in the browser

For example to proxy all request to `http://localhost:8100/api` to `http://my.server.com/api`

I hope this is useful for everyone..

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,30 @@ Use `ionic serve` to start a local development server for app dev and testing. T
 $ ionic serve [options]
 ```
 
+__Proxies:__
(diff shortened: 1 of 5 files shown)


axsauze:
This doesn't seem to work for query parameters..

keithdmoore:
You don't specify query params as part of defining the proxy

engehabahmed:
How to do that? should I change files beside ionic.project?

PR #136, opened 2014-11-06
feat(serve): Project Level configurable DocumentRoot

guilleferrer:
When working with Ionic projects I develop everything under a "src" directory instead of "www" and later I created my "gulp dist" task which concats, uglifies, minifies etc and places everything in the "www" folder.

That works perfectly but In order to use ionic serve and avoid executing my "gulp dist" task in the "watch"  I need "ionic serve" to use a different documentRoot, in my case, "src".

Then, my project configuration may change the documentRoot :
```
// ionic.project
{
  "documentRoot" : "src"
}
```


jbavari:
@guilleferrer - I'll be taking a look at this over the next week. Thanks for submitting it!

jbavari:
@guilleferrer - thanks for the submission. I've merged this into the 1.3.3 branch and should roll out very soon.

guilleferrer:
great ! :)

PR #196, opened 2015-01-20
remove upload promise

tlancina:
The promise wasn't really doing anything, since `ionic.fail` calls `process.exit(1)` anyways.  This fixes the response occasionally not emitting a 'data' event, and never emitting the 'end' event.


jbavari:
This will be rolling out in 1.3.3. Its in commit 178f702

PR #212, opened 2015-02-01
Fix displaying starter template list

zoso10:
When listing the starter templates with `ionic start --list` it would try to write the response to a file. For Linux and Mac, npm packages are installed under `/usr/lib/node_modules` or `/usr/lib/local/node_modules` which require root privileges to write files; so this would fail and not display the list. This fix just parses the response JSON.
I also updated the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,12 @@ __Local directory starters:__
 
 __Command-line flags/options:__
 
-    --appname, -a  .......  Human readable name for the app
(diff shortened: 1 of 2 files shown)


jbavari:
@zoso10  - thanks for the update. I'll be pushing this out in the next release. I've merged it with my next branch to go out, so i'll close this for now! Thanks again!

PR #221, opened 2015-02-04
Update README

jvannistelrooy:
The ionic serve flag `--labs` should be `--lab`


jbavari:
Thanks for the PR @jvannistelrooy!

PR #223, opened 2015-02-04
Fix for livereload on IP based addresses

duellsy:
postToLiveReload was assuming that the server was running localhost, this allows it to run on an IP based address


jbavari:
I'll be looking at this soon. Thanks for the submission!

PR #228, opened 2015-02-07
Suggest changing file permissions on OSX

gnomeontherun:
Inform users who don't want to (because you probably shouldn't) use `sudo` to install node modules globally how to update their file permissions.


jbavari:
Thanks for the PR @gnomeontherun!

PR #253, opened 2015-02-11
Use console.log instead of deprecated sys.puts

zarko-tg:
This prevents unnecessary warning coming from the deprecation of sys/util puts (See nodejs.org/api/util.html#util_util_puts)


jbavari:
I should be including this in the [https://github.com/driftyco/ionic-cli/blob/eccba8557ce3a3b9d873e39f2101cf9064464dba/lib/hooks/after_platform_add/010_install_plugins.js] latest hook-fixes branch.

PR #256, opened 2015-02-12
Hook fixes - 'ionic upload' should use documentRoot if set.

Drozes:
'ionic upload' was fixed to use the documentRoot if configured in
ionic.project instead only defaulting to the 'www' folder.

This should make it more inline with the server.js feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### Next-release
+
+* Updating task order in the CLI output for help - putting more important tasks at the top, and lesser used ones at the bottom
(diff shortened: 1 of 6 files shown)


jbavari:
@Drozes I'm not really sure what you are proposing here?

It looks like you are just trying to merge one of our branches to master, which will happen when its complete.

If this was not your intention, and you've got another issue, please open a valid issue describing the problem. Thanks!

Drozes:
I made the pole to link to my fork.

Wasn't sure how to contribute.  Get project - I love ionic!

PR #270, opened 2015-02-26
Allow the creation of the root directory based on an ionic.project file ...

vasconcelloslf:
Hello.

I've built with gulp an asset pipeline feature that runs through the ```ionic serve``` command. This pipeline takes the source files from ```my-project/www```, processes them in many different ways, then outputs the result in the ```my-project/www/dist``` dir, which contains the raw js/css that will be served trough ```ionic serve``` and wrapped by cordova in an android/ios project.

Here is my problem: since the content of the ```my-project/www/dist``` always comes from the same source (www), I don't track it with git, just like no one tracks the ```my-project/platforms``` dir. Because of that, after cheking out my project, the ```ionic serve``` throws an error and tells me that the ```www``` dir is not found. So this PR do the following:

- fix the ```ionic serve```error message to output the current dir used as the documentRoot
- allows the creation of the document root, if it doesn't exist, through a ```createDocumentRoot``` option in the ionic.project file

What do you think?

### DIFF
--- a/lib/ionic/serve.js
+++ b/lib/ionic/serve.js
@@ -75,6 +75,7 @@ IonicTask.prototype.loadSettings = function(cb) {
   this.printServerLogs = argv.serverlogs || argv['server-logs'] || argv.s;
   this.isAddressCmd = argv._[0].toLowerCase() == 'address';
   this.documentRoot = project.get('documentRoot') || 'www';
+  this.createDocumentRoot = project.get('createDocumentRoot');
   this.contentSrc = path.join(this.documentRoot, this.ionic.getContentSrc());
 
   this.getAddress(cb);
@@ -151,7 +152,11 @@ IonicTask.prototype.start = function(ionic) {
     var childProcess = null;
 
     if (!fs.existsSync( path.resolve(this.documentRoot) )) {
-      return ionic.fail('"www" directory cannot be found. Please make sure the working directory is an Ionic project.');
+      if (this.createDocumentRoot) {
+        fs.mkdirSync(this.documentRoot);
+      } else {
+        return ionic.fail('"' + this.documentRoot + '" directory cannot be found. Please make sure the working directory is an Ionic project.');
+      }
     }
 
     // gulpStartupTasks should be an array of tasks set in the project config


jbavari:
@vasconcelloslf Let me look over it and get back to you. Thanks for the submission.

jbavari:
@vasconcelloslf - this will be rolling out in 1.3.13. I had to merge it manually. Thanks again for the PR.

vasconcelloslf:
Awesome! Tks

vasconcelloslf:
@jbavari 

Speaking of build systems. I spent a solid week improving mine, and I would be happy to contribute to ionic in this direction.

I basically build the same asset pipeline feature from rails with gulp (with 1 or 2 more things). Is ionic interested in having this feature? If it is, I'm wiling to begin the work in a remote branch and follow any guideline or feedback that you guys may provide.

jbavari:
@vasconcelloslf of course. I've tried tackling the minification, but have had troubles trying to please every user in the direction of what tools to use.

Everyone has their own opinion about build systems. However, for those who don't we'd love your help with direction on this.

Do you have a current branch going you'd want me to take a look at?

vasconcelloslf:
@jbavari 

Hey, now I do. Check it ou the gulpfile https://github.com/vasconcelloslf/ionic-starter-app/blob/master/gulpfile.js

The idea is simple: I have a dev pipeline that runs with ```ionic start``` trough gulpStartupTasks and a build pipeline that runs with ```ionic build xxx``` through a cordova hook.

I use several gulp plugins to preprocess, minify, concat and replace constants (check https://github.com/vasconcelloslf/ionic-starter-app#8-configure-the-constant-values-gulp).

The dev pipeline doesn't concat nor minifies, and build pipeline does. I would love to work on it to incorporate this build system into ionic, probably as an optional feature.

I want to start a PR to begin to work, but first I would like to hear your feedbacks on the above build system.

jbavari:
@vasconcelloslf - Mike and I have taken a quick look at it.

I'll dig more into it today and give some feedback. Thanks again for the work.

vasconcelloslf:
@jbavari Ok, I will be waiting. Let me know if you need anything.

jbavari:
@vasconcelloslf I promise I'm not stalling on you. This is next on my list! I thought I was going to get to it sooner, but I've been getting distracted.

vasconcelloslf:
@jbavari Great! You must have lots of things to do. Just let me know if I can help you in anyway.

keithdmoore:
@vasconcelloslf How about using `--release` to trigger the prod workflow (aka minfy, uglify, etc).  prod constants would be used when `--release` is used.  Otherwise, dev workflow(aka `ionic serve`) would only do the constant replacements.   What do you think?

jbavari:
I have a few thoughts - while keeping the mindset that the masses will use this. By no means do I think I have all the answers.

Do you all think gulp is the way to go?

How will we update the gulpfile going forward? (Say between versions of the CLI).

Could we work this into a command instead of leaning on gulp?

How about customization for the users? Can we have it where users customize to their needs instead of the cookie cutter build? Something like utilizing the `ionic.project` file for customizations which folders to process?

keithdmoore:
maybe a `--compress` option would be better so the user has more flexibility

EdL:
Hi @jbavari - I have left a rather lengthy comment on this post https://github.com/driftyco/ionic-cli/issues/168 summarising my thoughts on the build step. I didn't post it here as I didn;t feel it was directly linked to this pull request.

My view on the questions you posted above would be:

Do you all think gulp is the way to go?
Yes

How will we update the gulpfile going forward? (Say between versions of the CLI).

I guess the Gulpfile could just be updated and available via Git as per latest nightly builds. Then when you release CLI version, it is included in there.

Could we work this into a command instead of leaning on gulp?
I think Gulp seems to be the way most people have gone, it's popular, easy to extend or customise etc. It may also be a lot easier to use the Gulp plugins for the various tasks involved in the build.

How about customization for the users? Can we have it where users customize to their needs instead of the cookie cutter build? Something like utilizing the ionic.project file for customizations which folders to process?

Yes. Custom folder for source and destination would be good as well as users being able to add steps to the gulpfile if they wish. At least there will be a common starting point and hopefully it will do whatthe the vast majority needs. Using the ionic.project file seems like a good idea.

Thanks

softBarbarian:
@jbavari hello is there any update for this?

PR #274, opened 2015-02-27
Additional Proxy config parameter to allow connection pooling to be turned off

mur-dog:
I have been using the proxy feature to avoid CORS issues to our API.  API server is JHipster/Tomcat8.

However we have been running into an annoying problem - after 5 livereloads our application would stop working.  Upon debugging I tracked down the issue to the proxy hanging - it looks like for some reason the proxy is not properly releasing/marking the Http request as complete.  So the default HttpAgent spins up a new socket connection (up to 5) before getting into a blocked state.

This addition allows the proxy request to opt out of the default connection pooling by passing agent=false to the http request - and prevents the deadlock issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ In the `ionic.project` file you can add a property with an array of proxies you 
 
 * `path`: string that will be matched against the beginning of the incoming request URL.
 * `proxyUrl`: a string with the url of where the proxied request should go.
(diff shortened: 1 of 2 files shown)


jbavari:
Nice fix. I'll look at this today.

jbavari:
@mur-dog - due to some pretty big changes to serve.js, I had to merge this in manually. Thanks for the contribution!

PR #301, opened 2015-03-16
Update package.json

sazap10:
Removes the resolved url for open, instead uses the version.


jbavari:
@sazap10 As a heads up, going forward we've ripped out a lot of code from the CLI to put it into its own library. In this library, the code you propose has been merged. Thank you for the contribution!

PR #358, opened 2015-04-07
Remove cordova-plugin-crosswalk-webview from plugins on crosswalk revert

mattezell:
Removing 'cordova-plugin-crosswalk-webview' from plugins when running 'ionic browser revert android' as plugin is crosswalk resource and so should no longer remain after reverting to Android browser - presence of this plugin caused failure of subsequent attempts to run 'ionic platform add android' after removal of Android platform from project.


mattezell:
Note: corresponds with an issue I incorrectly opened in the 'ionic' project - https://github.com/driftyco/ionic/issues/3482

jbavari:
Thanks for the contribution!

mattezell:
Anytime!  Thank you guys for the continued hard work and investment into the ecosystem!

PR #404, opened 2015-04-28
add rejectUnauthorized option to the proxy configuration to allow untrusted certificates

nzamani:
See http://forum.ionicframework.com/t/how-to-use-a-proxy-to-https-services/21854/2

This change allows the proxy to accept untrusted certificates, i.e. when proxying to a server on localhost + https

### DIFF
--- a/lib/ionic/serve.js
+++ b/lib/ionic/serve.js
@@ -326,7 +326,9 @@ IonicTask.prototype.start = function(ionic) {
         var opts = url.parse(proxy.proxyUrl);
         if(proxy.proxyNoAgent)
             opts.agent = false;
-
+        
+        opts.rejectUnauthorized = !(proxy.rejectUnauthorized === false);
+        
         app.use(proxy.path, proxyMiddleware(opts));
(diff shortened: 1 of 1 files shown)


jbavari:
@nzamani As a heads up, going forward we've ripped out a lot of code from the CLI to put it into its own library. In this library, the code you propose has been merged. Thank you for the contribution!

nzamani:
@jbavari Good to hear, Thanks! Do you already know when and in what version this feature will be available? The reason I ask is because I wanna make sure that our "local" change does not disappear by an update of ionic-cli that does not include the change...

nzamani:
Just in case anyone is interested:
It seems that after the refactoring the feature has moved from ionic-cli to ionic-app-lib. There you can check lib/serve.js, see here:

https://github.com/driftyco/ionic-app-lib/blob/2dece25d83d9149d900766bd5dc6d1e5780a1091/lib/serve.js#L317

or https://github.com/driftyco/ionic-app-lib/blob/master/lib/serve.js#L317

jbavari:
@nzamani Thanks for sharing that knowledge and those links. This will be helpful.

I'm also updating the wiki to give instructions on how/why it is laid out this way.

PR #465, opened 2015-06-05
Windows platform on lab

CKGrafico:
Add windows to the lab #464


jbavari:
At this current time, the preview is too wide for 3 views. 

We'll address this and redesign the layout so you can select which views to show. 

Thank you for the pull request.

PR #517, opened 2015-07-06
Avoid race condition when installing plugins

ReactiveRaven:
When installing multiple plugins asyncronously, a race condition can cause plugins to fail to install, or fail to be registered as installed. As an example, asyncronously installing multiple (5+) plugins often causes `plugins/{platform}.json` to drop plugins as they are all trying to write to it at once.


cantecim:
It's great. it must be merged asap

jbavari:
@ReactiveRaven I'm going to have to manually merge this one. Hang tight.

Freundschaft:
is this still a bug actually?

jthoms1:
Thanks for opening the issue. It seems that issue has been resolved. I am going to close this issue but please feel free to reopen another issue if this occurs again.

PR #520, opened 2015-07-07
ionic -v now also reports version of the current project (if applicable)

toddhalfpenny:
`ionic -v` and `ionic --version` now also return the name and version of the current project (if there's a package.json in the current directory.

Example repsonse is
```
Ionic: 1.6.1
Current project (MyLovelyAppName) : 0.1.1
```

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -423,7 +423,15 @@ Cli.processExit = function processExit(code) {
 };
 
 Cli.version = function version() {
-  console.log(settings.version + '\n');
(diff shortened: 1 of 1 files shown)


jbavari:
After discussing with the team, the `ionic info` command handles exactly what you're asking for, and with a lot less effort to get it. The `ionic -v` is following most CLI conventions of simply providing the version only.

For now, we've decided to put this one on the back burner and focus on more pressing matters.

Thank you for the contribution.

PR #523, opened 2015-07-11
encoding options

cantecim:
fs.readFileSync and fs.writeFileSync encoding options passed with an object

There was a bug in non-latin characters that will convert them to html entities


jbavari:
@cantecim Thanks! We appreciate your PR.

PR #566, opened 2015-08-15
Specify a different address for devices to connect to for ionic run --livereload

lewisl9029:
I'm running ionic-cli from a Docker container that has a different IP than the host machine's IP on the LAN. `ionic run --livereload` by default builds an app that attempts to connect to the IP it serves on (the Docker container's IP in this case, which does not exist on the LAN).

This PR adds a `--external-address` parameter where users can specify a different address for the device to connect to.

I pushed a version to https://www.npmjs.com/package/ionic-run-external-address to use for my app in the mean time. Feel free to give it a spin if you need something like this now.

Let me know if there's anything that needs to be changed.

Related issue: #557

### DIFF
--- a/lib/ionic/cordova.js
+++ b/lib/ionic/cordova.js
@@ -261,6 +261,9 @@ IonicTask.prototype.setupLiveReload = function(argv) {
   }
 
   var externalAddress = argv['external-address'] || argv.e;
+  if (externalAddress) {
+    console.log('Connecting to ' + externalAddress + ' from device');
+  }
 
   return promise
   .then(function() {
--- a/lib/ionic/cordova.js
+++ b/lib/ionic/cordova.js
@@ -136,12 +136,14 @@ IonicTask.prototype.runCordova = function(cmdName, argv) {
   var cleanArgs = [];
   var port = argv.port || argv.p || '';
   var liveReloadPort = argv.livereloadport || argv['livereload-port'] || argv.r || '';
-  var ignoreCmds = '--livereload -l --consolelogs -c --serverlogs -s --port -p --livereload-port -i -r'.split(' ');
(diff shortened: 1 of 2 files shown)


jbavari:
@lewisl9029 I like what you've done here.

A few questions - have you tried the `--all` option? That opens it up to use the address `0.0.0.0`. 

Also there's the `--address` option to specify an address. Does this not satisfy your needs?

lewisl9029:
Yep, I've tried both approaches but neither of them work for my use case.

--all: `ionic run --livereload --all` will build a Cordova app that attempts to connect to 0.0.0.0 from the device, and 0.0.0.0 does not exist as a real address on the LAN, so the connection would fail.

--address: If for instance my host machine has an address of 192.168.1.3, running `ionic run --livereload --address 192.168.1.3` from within a docker container will bring up a dev server that's unreachable, because the address 192.168.1.3 doesn't map to a valid network interface inside the docker container, which has different network configuration than the host.

The --external-address flag in this PR allows `ionic run --livereload` to serve on a valid address inside the container, but build a Cordova app that would try to connect to a different, reachable address on the LAN.

Let me know if you have any more questions or concerns.

AgDude:
@lewisl9029 I would love to see this functionality working. What network configuration are you expecting for the VM? From your description I don't understand how the device will reach the actual live-reload server. Are you opening a firewall port on the host machine, to a port forwarded from the VM?

lewisl9029:
@AgDude Yep that's exactly what I'm doing.

The container and VM already had port forwarding configured to the point where you can access the regular `ionic serve` dev server from the host machine as localhost. All I had to do to make the server reachable from LAN is open the forwarded ports on the host machine's firewall.

jerouris:
Will this pull request get merged in master? This is functionality that is very usefull when you are running from a VM.

PR #590, opened 2015-09-20
Tasks multi-language

CarmaZone:
* Now ionic-cli helper is multi-language.
* Tasks traslated to spanish

<strong>How to use:</strong>

Only write the language argument (*--es* to spanish)
Example: **ionic -h --es**
tasks_to_spanish

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -9,6 +9,7 @@ var Cli = module.exports,
     path = require('path'),
     settings = require('../package.json'),
     Tasks = require('./tasks/cliTasks'),
+    Tasks_es = require('./tasks/cliTasks_es'),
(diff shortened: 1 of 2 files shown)


jbavari:
@CarmaZone I'd love to tackle this.

Currently, with the Ionic CLI v2 rolling out, we'll have some tasks changing quite a bit. This change is duplicating the cliTasks.js file for translation.

Thinking going forward, each time we modify the cli tasks, we'll need to modify the associated translated version. Would there be anyway to do this without duplicating those tasks?

Put another way, how would we add another language? Each time we modify tasks, we'd have to update all. 

Could we tackle the problem with having a translation, so to speak, that reads the tasks, then pulls the associated language if another language is passed?

PR #631, opened 2015-10-28
Feat(start): dynamically load v2 or v1 start task from app lib

tlancina:
Depends on https://github.com/driftyco/ionic-app-lib/pull/38


jbavari:
Also solid. Lets merge it to v2.

PR #830, opened 2016-03-10
Update bower error msg

Fuiste:
Stating that the package couldn't be found on any bower error is confusing when things like version conflicts occur (This happens a _lot_ with bower, especially when adding the `ionic-platform-web-client`).

Ideally, we should make bower a dependency of the cli, and not be installing via `exec('some-command-here')`, but this should at least provide a bit better info.


tlancina:
:+1: lgtm

PR #888, opened 2016-04-01
Support for Typescript Generators

andrewmcgivery:
This adds support to pass a --ts argument to `ionic g` to generate typescript code instead of ES6.

Supported by driftyco/ionic#6008


tlancina:
Awesome, thanks for the PR!

PR #926, opened 2016-04-11
Add testing process

jthoms1:
This pull request will add a few testing features around cirleci integration. This includes:
- Istanbul and Coveralls integration
- CircleCI testing on multiple supported versions of Nodejs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.cookies
 node_modules/
 .idea/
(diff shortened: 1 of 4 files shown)


tlancina:
We don't support node 0.10, so I think you can remove it (officially I think we only support > 4, but a lot of people still use 0.12, so good to test).

Other than that, lgtm.

PR #949, opened 2016-04-20
feat(stats): update stats to track --v2 and --ts flags; add os, gulp,…

jthoms1:
Added code so that we will track information for v2. Also included ability to get metrics on nodejs, gulp and os versions.


tlancina:
Looks great, we may want to add the `--skip-npm` flag at some point as well if that is of interest now that `start` does npm install by default.

PR #1015, opened 2016-05-20
Fix #1014

### LINKED ISSUE #1014: Cli arguments are not passed to cordova
Ionic cli wraps cordova cli and removes its params before calling cordova, but the '--address' special case is not well handled, and when you have a ionic arg skipped, all next valid args are skipped too

you can add any params to cordova cli as you can use them in hooks

then:
```
ionic run --device --livereload --consolelogs --myparam
```
will launch 
```
cordova run --device
```
instead of
```
cordova run --device --myparam
```
i will provide a PR to fix that

jetwitaussi:
Send arguments from ionic cli to cordova cli


jthoms1:
This PR is no longer applicable with the latest refactor.  Sorry it took so long to get around to this, but we do appreciate the PR!

Thanks again!

PR #1091, opened 2016-06-20
fix(#594): propagate Cordova error code

### LINKED ISSUE #594: bug: Failing hook does not fail build (exit code)
_From @NicholasTuck on September 23, 2015 19:50_

**Type**: <span ionic-type>bug</span>

**Platform**: <span ionic-platform>all</span>  

<span ionic-description>If I have a failing hook that exits non 0 (like a before_build doing a webpack build), ionic will recognize the hook failed, and will stop building, but it exits with code 0. This makes my automated build tool, Jenkins, not know it's actually failing.  

cordova build fails in a nearly identical way, but it does exit code 1 as expected.

Ionic reporting exit 0: 
image

Cordova reporting exit 1: 
image
</span>

<span is-issue-template></span>

_Copied from original issue: driftyco/ionic#4417_

jakub-g:
Fix #594 

Hello, I believe that aborting and exiting right away in case of Cordova command failure is a sensible way to do, but I don't know the whole ionic codebase, so I'm open for suggestions from the ionic team.

FYI when I `throw` instead, the ionic process still exit with code `0`, that's why I used `process.exit()` directly.

To reproduce the bug clone https://github.com/jakub-g/ionic-cli-bug-594 and follow README

To reproduce the fix, `npm uninstall -g ionic`, clone this branch, `npm link` and later follow README as above

### DIFF
--- a/lib/ionic/cordova.js
+++ b/lib/ionic/cordova.js
@@ -72,8 +72,12 @@ IonicTask.prototype.run = function(ionic, argv) {
       return self.runCordova(cmdName, argv);
     })
     .then(function(runCode) {
-      //We dont want to do anything if the cordova command failed
-      if(runCode !== 0 || argv.nosave) {
+      // If the cordova command failed, abort immediately and propagate the error code
+      if (runCode !== 0) {
+        console.log(('Cordova command failed with error code ' + runCode + ', aborting!').red.bold);
+        process.exit(runCode);
+      }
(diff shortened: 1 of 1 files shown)


jthoms1:
Thanks for opening the pull request. It seems that issue has been resolved. I am going to close this issue but please feel free to reopen another issue if this occurs again.

PR #1299, opened 2016-08-21
Add watch for ionic emulate

sagivf:
This is needed for `ionic emulate ios` to live reload on css changes


sagivf:
https://github.com/driftyco/ionic-cli/issues/1276#issuecomment-241260762

jthoms1:
This code has been moved into @ionic/app-scripts and the process is no longer in gulp.

PR #1411, opened 2016-09-25
fix(generators): allow configurable directory to be passed in

danbucholtz:


ramonornela:
After merge think important the docs with help ;).

PR #1415, opened 2016-09-27
docs(start): add options wrapper e starter branch

ramonornela:
Reopen pull #1392, I think that now is the moment  😉 version 2.1.0 released.


jthoms1:
Still have the same opinion here on documenting these.

PR #1614, opened 2016-11-02
feat(): allow emulate and run to use serve for livereload.

jthoms1:


jthoms1:
This is dependent on the next release of app-lib

vejja:
Hi 
I am using ionic-cli v2.1.7 and ionic-app-scripts v0.0.42, so I believe feat #1614 should be enabled.

1. `--livereload`watches for changes in /www but not in /src. For this we would need the `emulate` command to use `ionic-app-script watch` under the hood instead of `ionic-app-script build` 
2. passing `--dev` or `--debug` has no effect. It would be great to be able to pass this as argument to both the `run` and `emulate` commands, and even greater to have it as default for `emulate` (similar to what the `serve` command does) to allow faster rebuilds
3. Not sure where this comes from but I get the following warning :
```
(node:24385) DeprecationWarning: Using Buffer without `new` will soon stop working. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.
```

PR #1716, opened 2016-11-23
Allow specifying of buildId

superkew:
This will allow a user to use the following command:
ionic package download --buildId=50


superkew:
Is there anything I need to do to get this thing approved?

PR #1744, opened 2016-12-01
Fix `ionic lib update` "invalid version: undefined" error

flanger001:
This switches the arguments in `getVersionData` and stops the **`Invalid version: undefined`** error when users run `ionic lib update`.

### DIFF
--- a/lib/ionic/lib.js
+++ b/lib/ionic/lib.js
@@ -98,7 +98,7 @@ function printLibVersions(data) {
 }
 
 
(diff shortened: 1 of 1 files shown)


flanger001:
I see the tests fail but there's nothing actually testing this function in the first place. I am not exactly sure how to test it myself.

amiral84:
I get the same error when I try to update the ionic to version 1.0.0.

PR #1747, opened 2016-12-04
fix #1183: run before-hooks with locally installed gulp#4.0

### LINKED ISSUE #1183: use gulp#4
I was just trying to use gulp 4 instead of 3 in the gulpfile the `ionic start` command generates, but it looks like your CLI is dependent on version 3 and it cannot find some tasks.

Note that using v4 you get the advantage of dropping some deps, such as `run-sequence`, since v4 already provides a way to run things in series or parallel.

I get:

``` bash
TypeError: Cannot read property 'serve:before' of undefined
    at Object.runWithGulp (/Users/rolandjitsu/Projects/infarm/operator/node_modules/ionic/lib/cli.js:217:17)
    at Object.run (/Users/rolandjitsu/Projects/infarm/operator/node_modules/ionic/lib/cli.js:190:16)
    at Object.<anonymous> (/Users/rolandjitsu/Projects/infarm/operator/node_modules/ionic/bin/ionic:13:10)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.runMain (module.js:575:10)
    at run (bootstrap_node.js:352:7)
```

I could submit a pr if this would be an option, should be a straight forward migration.

antonanders:
gulp-4.0 changes the way tasks are registered and run internaly, so ionic-cli does not work when you install gulp 4.0. These changes will make before-hooks work again.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -228,13 +228,21 @@ Cli.runWithGulp = function runWithGulp(argv, taskInstance, rawCliArguments) {
     log.error(chalk.red('Do you need to run `npm install`?\n'));
     return process.exit(1);
   }
+  
+  // workaround for local gulp version >= 4.0
+  var gulpTasks = gulp.tasks || {};
+  if (gulp['tree'] && gulp.tree().nodes) {
+    gulp.tree().nodes.forEach(function (task) {
+      gulpTasks[task] = true;
+    })
+  }
 
   // setup gulp logging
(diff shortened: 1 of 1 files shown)


brunodeprez:
Please include in next release, thanks!

PR #1804, opened 2016-12-22
feat(start): Yarn option to install dependencies

jsayol:
Add an option to install Node dependencies via yarn instead of npm.

Goes hand-in-hand with driftyco/ionic-app-lib#134


jsayol:
Closing. See: https://github.com/driftyco/ionic-app-lib/pull/134#issuecomment-268877814

PR #2019, opened 2017-03-21
Prompts

jgw96:
This is the cli part of the new generators for Ionic 3. This PR adds the following:
- New prompts for all the possible types of components you can generate
- Tie in with code splitting in Ionic 3
- The ability to prompt the user and from the answers decide exactly where to tell app scripts put a 
    pipe, provider or component to best work with code splitting


packages/ionic/src/commands/generate.ts
```diff
@@ -9,10 +9,12 @@ import {
 
 import { load } from '../lib/utils/commonjs-loader';
 
+import { prompt, tabsPrompt } from '../lib/generate';
+
 @CommandMetadata({
   name: 'generate',
   aliases: ['g'],
-  description: 'Generate pages and components',
+  description: 'Generates pipes, components, pages, directives and tabs',
```
dwieeb: Add comma after directives, please

packages/ionic/src/commands/generate.ts
```diff
@@ -40,14 +42,67 @@ export class GenerateCommand extends Command {
...
+    let context;
+
     process.argv = ['node', 'appscripts'];
     const appScripts = load('@ionic/app-scripts');
-    const context = appScripts.generateContext();
 
-    tasks.next('Generating');
+    switch (type) {
+      case 'page':
+        context = appScripts.generateContext();
```
dwieeb: context can be generated once, above this switch statement

packages/ionic/src/commands/generate.ts
```diff
@@ -40,14 +42,67 @@ export class GenerateCommand extends Command {
     const [ type, name ] = inputs;
     const tasks = new TaskChain();
 
+    let context;
+
     process.argv = ['node', 'appscripts'];
```
dwieeb: there is a new way to do this, talk to @jthoms1

packages/ionic/src/lib/generate.ts
```diff
@@ -0,0 +1,66 @@
+import * as path from 'path';
+
+
+export async function getPages(appScripts: any, context: any) {
+  const fileChoices: any[] = [];
+
+  const pages = await appScripts.getNgModules(context, ['page', 'component']);
+
+  pages.forEach((page: any) => {
+    fileChoices.push({ fileName: path.basename(page.absolutePath, '.module.ts'), absolutePath: page.absolutePath, relativePath: path.relative(context.rootDir, page.absolutePath) });
```
dwieeb: `.module.ts` shouldn't be hard coded. app scripts has a config value for it

packages/ionic/src/lib/generate.ts
```diff
@@ -0,0 +1,66 @@
+import * as path from 'path';
+
+
+export async function getPages(appScripts: any, context: any) {
+  const fileChoices: any[] = [];
+
+  const pages = await appScripts.getNgModules(context, ['page', 'component']);
+
+  pages.forEach((page: any) => {
```
dwieeb: `.map` would be cooler to use here

packages/ionic/src/lib/generate.ts
```diff
@@ -0,0 +1,66 @@
...
+    fileChoices.push({ fileName: path.basename(page.absolutePath, '.module.ts'), absolutePath: page.absolutePath, relativePath: path.relative(context.rootDir, page.absolutePath) });
+  });
+
+  return fileChoices;
+}
+
+export async function prompt(name: string, appScripts: any, context: any, inquirer: any) {
```
dwieeb: `name` maybe should be `type`?

packages/ionic/src/lib/__tests__/generate.ts
```diff
@@ -0,0 +1,66 @@
...
+    const inquirer = {
+      prompt: jest.fn(() => {
+        return {
+          usage: true
+        };
+      })
+    };
+
+    const result = await tabsPrompt({}, inquirer);
+    expect(result).toEqual(jasmine.any(Array));
```
dwieeb: you can be more specific here, can't you? any array could be empty, have weird values, etc.

packages/ionic/src/lib/generate.ts
```diff
@@ -0,0 +1,66 @@
...
+    type: 'confirm',
+    name: 'usage',
+    message: `Use this ${name} in more than one template?`
+  });
+
+  if (!usageQuestion.usage) {
+    const filteredChoices: any = [];
+    const fileChoices = await getPages(appScripts, context);
+
+    fileChoices.forEach((file: any) => {
```
dwieeb: `.map` ? see right below

packages/ionic/src/lib/__tests__/generate.ts
```diff
@@ -0,0 +1,66 @@
+import { prompt, getPages, tabsPrompt } from '../generate';
+
+describe('prompt', () => {
+
+  it('should return a file path', async () => {
+    const inquirer = {
+      prompt: jest.fn(() => {
```
dwieeb: one reason one might do `jest.fn(fn)` is to use the jest mock utilities to check which calls occurred: https://facebook.github.io/jest/docs/mock-function-api.html

Issue #2058, opened 2017-04-02
Server settings ipAddress undefined in emulate command


twinj:
#### What version of the CLI are you using? Output of `ionic --version`
3.0.0-beta.5

#### Short description of the problem:
When running the emulate live reload command the content src in the config.xml is being set as

> <content src="http://undefined:8100" original-src="index.html"/>

#### What behavior are you expecting?

I am expecting to see a proper IP address.

>   <content src="http://28.28.28.30:8100" original-src="index.html"/>

**Steps to reproduce:**
1. `ionic cordova:emulate --target="iPad-Air-2, 10.3" --livereload`

```
serverSettings.ipAddress is undefined causing the config xml to look like this

<content src="http://undefined:8100" original-src="index.html"/>
```

**Post the output of `ionic info` below please**
 Cordova CLI: 6.5.0 
    Ionic CLI Version: 3.0.0-beta.5
    ios-deploy version: 1.9.1 
    ios-sim version: 5.0.13 
    OS: macOS Sierra
    Node Version: v6.10.1
    Xcode version: Xcode 8.3 Build version 8E162

**Other information:** (e.g. stacktraces, related issues, suggestions how to fix, stackoverflow links, forum links, etc)

jthoms1:
Could you provide me with the output of the `cordova:emulate` command?

twinj:
I was meant to add more details but got side tracked after creating the pull release which fixes this.

The command works and deploys the application however when the index page is loaded it looks for the content src it looks like this.

` <content src="http://undefined:8100" original-src="index.html"/>`

twinj:
This has been fixed

PR #2059, opened 2017-04-02
Fix live reload on emulator - change emulate ipAddress to publicIp

twinj:
Server settings ipAddress undefined in emulate command, changed to be same as run command.


twinj:
Fixes #2058

PR #2128, opened 2017-04-21
Further prioritize current version in README

piotrowski:
I further prioritized the current version instructions in the README by removing the "v2" with "current version" and moving the v1 stuff to the back.


dwieeb:
Thank you, @piotrowski !

PR #2217, opened 2017-05-05
Pass chosen ip from list back to options

### LINKED ISSUE #2082: Make chosen IP available in app scripts
**Note: for support questions, please use one of these channels:**

#### What version of the CLI are you using? Output of `ionic --version`
3 beta 5

#### Short description of the problem:
When running a live reload build I would like to be able to access the chosen Ip through the app scripts in my custom web pack config so that I can use it in my scripts, e.g: CORS policy changes in the index.html for different machines and environments.

`ionic cordova:emulate --target="iPad-Air-2, 10.3" --debug --livereload --webpack="custom"`

Currently the chosen ip is only passed to the caller of serve and not serve itself. The ionic cli then modifies the config.xml using that IP but does not make it available.

The chosen IP could be added to the CommandLineOptions object using the 'address' key.

twinj:
This enables use of the chosen address from the app-scripts

Fixes #2082

### DIFF
--- a/packages/cli-plugin-ionic-angular/src/serve.ts
+++ b/packages/cli-plugin-ionic-angular/src/serve.ts
@@ -32,6 +32,7 @@ export async function serve(args: CommandHookArgs): Promise<{ [key: string]: any
           choices: availableIPs.map(ip => ip.address)
         });
         chosenIP = promptAnswers['ip'];
+        args.options.address = chosenIP;
       }
     }
   }


jthoms1:
This would cause some issues with ionic-app-scripts.  Currently ionic-app-scripts expects `--address` to be passed from command line.  App-scripts is a process that we interact with from the cli-ionic-angular plugin, but in doing so we are honoring the same command line arguments that it expects.  Introducing this may not seem like it would cause a problem, but it would be breaking the contract that we have.

I appreciate the time that you have put into this issue, but you may be better off calling wrapping the ionic cli command with a node script and passing the ip address as an argument in that way, then you would not need to to hardcode the values.

Thanks again!

twinj:
From my take of the code App scripts serve is called from within this plugin and all options are passed to it as given from the cli options. However if the address is chosen then that option should also be passed. Address could be defined using an environment variable, a package config variable, or by the command line. This is true for all ionic app script options. So why pass any options at all as this would break the contract for other options. Why is address so different?

dwieeb:
See https://github.com/driftyco/ionic-cli/issues/2082

dwieeb:
@twinj Just updating this to let you know that I am aware of the issue here. I've been trying to find a chunk of time to sit down, understand it, and figure out a good solution to this. I just haven't been able to due to the upcoming release, and I apologize.

This may need to wait until post-release.

dwieeb:
@twinj This was automatically closed because the `v3` branch was deleted. You should be able to remake it now on `master`. Sorry about that.

PR #2235, opened 2017-05-09
core(docs): adding `ionic` to command name

perrygovier:


dwieeb:
@perrygovier We shouldn't even need to commit the docs to this repo, right?

perrygovier:
no, the change to `scripts/createMarkdownDocs.ts` is still necessary, but the entire `docs` directory should be safe to ignore

dwieeb:
merged--not sure why github can't detect this

PR #2239, opened 2017-05-10
fix(readme): remove extra @latest in npm install

eikooc:
Makes the npm installer install the stable build instead of being on the @latest branch


dwieeb:
Thanks! The `@latest` dist-tag is redundant.

PR #2326, opened 2017-05-21
Added check to verify if name is empty at generate.

Bernardoow:
I added new check for empty names.


dwieeb:
@Bernardoow Perfect! Thank you so much.

PR #2345, opened 2017-05-25
feature(config): --timeout / --no-timeout flags

### LINKED ISSUE #2337: --timeout / --no-timeout cli flags
The timeouts are a good idea for people when a network hiccup happens so they aren't waiting forever while it hangs. But for some people on really slow connections, `--no-timeout` is a good thing to implement.

ricardo-mello:
Closes #2337


dwieeb:
This looks perfect! Great work.

PR #2359, opened 2017-05-28
chore(package): bump dev dependencies

danielsogl:
Update the dependencies to the latest feature and bugfix releases without breaking changes


dwieeb:
Thanks

PR #2368, opened 2017-05-30
refactor(config): set request timeout instead of deadline timeout.

ricardo-mello:
This changes applies a `request timeout` that waits for server response, instead of the default `deadline timeout` that waits for the entire thread.

### DIFF
--- a/packages/cli-plugin-cordova/src/lib/__tests__/resources.ts
+++ b/packages/cli-plugin-cordova/src/lib/__tests__/resources.ts
@@ -303,7 +303,7 @@ describe('@ionic/cli-plugin-cordova', () => {
         imageId: '60278b0fa1d5abf43d07c5ae0f8a0b41'
       }];
 
-      const response = await resources.uploadSourceImages(sourceImages, 0);
+      const response = await resources.uploadSourceImages(sourceImages, false);
       expect(response).toEqual([{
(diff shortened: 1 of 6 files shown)


types/superagent.d.ts
```diff
@@ -118,7 +118,7 @@ declare module "superagent" {
         send(): this;
         set(field: string, val: string): this;
         set(field: Object): this;
-        timeout(ms: number): this;
+        timeout(ms: number | Object | { response: number, read: number, deadline: number }): this;
```
dwieeb: Why also union this type on `Object` ?

packages/cli-plugin-cordova/src/lib/__tests__/resources.ts
```diff
@@ -303,7 +303,7 @@ describe('@ionic/cli-plugin-cordova', () => {
         imageId: '60278b0fa1d5abf43d07c5ae0f8a0b41'
       }];
 
-      const response = await resources.uploadSourceImages(sourceImages, 0);
+      const response = await resources.uploadSourceImages(sourceImages, false);
```
dwieeb: It's okay to have the 2 min `response` timeout for these, too, right?

dwieeb:
Merging as per our discussion in slack

Issue #2431, opened 2017-06-14
--labs does not pass config.get() values defined with ?ionicsomeparam=somevalue


mmolhoek:
**Description:**
It seems that, when you start ionic in labs mode with --lab, you do not have the possibility to pass config options by extending the url with ?ionicsomesetting=1, which you would normally get in your app with config.get('somesetting').

It works fine without labs mode, but not within labs mode. i can see in the chrome inspector that the parameters are not passed to the individual iframe(s) of the specific platform(s) you selected in labs.

**My `ionic info`:**

global packages:

    @ionic/cli-utils : 1.4.0
    Cordova CLI      : 7.0.1
    Ionic CLI        : 3.4.0

local packages:

    @ionic/app-scripts              : 1.3.7
    @ionic/cli-plugin-cordova       : 1.4.0
    @ionic/cli-plugin-ionic-angular : 1.3.1
    Cordova Platforms               : android ios 4.4.0
    Ionic Framework                 : ionic-angular 3.3.0

System:

    Node       : v7.5.0
    OS         : macOS Sierra
    Xcode      : Xcode 8.3.3 Build version 8E3004b
    ios-deploy : 1.9.1
    ios-sim    : not installed
    npm        : 4.1.2

**Other Information:**

mmolhoek:
just created PR https://github.com/ionic-team/ionic-cli/pull/2438 to fix this

PR #2438, opened 2017-06-15
patch for passing params from lab to iframes

mmolhoek:
this little pull request makes the labs (-lab) environment pass all params put on the url to the iframes that run the iphone/android and windows ionic.

this way you can pass ionic params to your app also when running in lab mode.

fix for https://github.com/ionic-team/ionic-cli/issues/2431


dwieeb:
Thanks @mmolhoek ! Looks good.

PR #2446, opened 2017-06-16
Add [details] tag around `ionic info`

janpio:
... so that posting in forum or issues is automatically collapsed.

(I'm not 100% sure if this is actually a good idea - you decide)


dwieeb:
Yeah... people post this in github, slack, all over the place. I contemplated having `ionic info` formatted in markdown, or maybe some `--format` flag, but decided against implementing for now.

PR #2475, opened 2017-06-26
chore(ionic-cli): bump strip-ansi

danielsogl:
No breaking changes


dwieeb:
We should be able to bump all of chalk. I'll take a look into this. This shouldn't be a peer dep, anyway.

Thanks for bringing it to my attention

PR #2509, opened 2017-07-10
fix(dependencies): remove peer deps since they are breaking apps

danbucholtz:


dwieeb:
I was fixing this locally this morning 👍

PR #2526, opened 2017-07-14
docs(readme): change example proxy

janpio:
Using example.org as intended in documentation.
Added port as I assume most proxies will use one.


dwieeb:
Thanks, good idea.

PR #2551, opened 2017-07-21
feat(generators): add module options

mhartington:
Adds `module` and `no-module` flags to the CLI

Requires https://github.com/ionic-team/ionic-app-scripts/pull/1141 to be merged first

### DIFF
--- a/packages/cli-plugin-ionic-angular/src/generate.ts
+++ b/packages/cli-plugin-ionic-angular/src/generate.ts
@@ -11,7 +11,7 @@ import {
 import * as AppScriptsType from '@ionic/app-scripts';
 
 import { load } from './lib/modules';
-import { prompt, tabsPrompt } from './utils/generate';
+import { getModules, prompt, tabsPrompt } from './utils/generate';
 
 export async function generate(args: CommandHookArgs): Promise<string[]> {
   if (!args.env.project.directory) {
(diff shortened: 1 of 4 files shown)


packages/cli-plugin-ionic-angular/src/generate.ts
```diff
@@ -37,21 +37,25 @@ export async function generate(args: CommandHookArgs): Promise<string[]> {
...
   const [ type, name ] = args.inputs;
+  const includeNgModule = args.options.module;
+  const includePageConstants = args.options.constants;
+
+  // args.env.log.msg(() => `${includePageConstants}`);
 
   switch (type) {
     case 'page':
-      await AppScripts.processPageRequest(context, name);
+      await AppScripts.processPageRequest(context, name, includeNgModule, includePageConstants);
```
dwieeb: Maybe move the options into an object? I can imagine there being a _lot_ more than 2 options for generators. We probably don't want 10+ positional args, especially cross-library.

packages/ionic/src/commands/generate.ts
```diff
@@ -35,6 +35,20 @@ The given ${chalk.green('name')} is normalized into an appropriate naming conven
       description: 'The name of the component being generated',
     }
   ],
+  options: [
```
dwieeb: You can provide examples of these options using the `exampleCommands` list above.

packages/cli-plugin-ionic-angular/src/generate.ts
mhartington: Makes sense

packages/ionic/src/commands/generate.ts
mhartington: it is done

PR #2553, opened 2017-07-24
docs(readme): make service proxy json more real

janpio:
CLI adds "type" attribute, "email" isn't normally present.


dwieeb:
Ah, thanks 👍

PR #2554, opened 2017-07-24
docs(readme): More information about Service Proxies

janpio:
added relationship to all livereload commands, not only `ionic serve`
added note about usage in app code
added note about required restart of command


dwieeb:
We could probably have these commands also watch `ionic.config.json` and reload it.

PR #2557, opened 2017-07-24
docs(readme): replace :heart:

janpio:
with :sparkling_heart: because it is rendered properly on Windows as well.
:heart: is sometimes quite strange, see diff:
image


dwieeb:
Can you PR Windows instead? 

😂

janpio:
Seriously, I thought about investigating **why** this is displayed so shitty for 3 seconds, then decided to just do this PR to get rid of it and be done with it ;)

dwieeb:
It's fine 😛 I like sparkles.

timbru31:
funniest PR ever 😂

dwieeb:
💖

PR #2589, opened 2017-08-01
Update command names

gfvcastro:


dwieeb:
@gfvcastro What issue does this solve?

Those command names are part of the `CordovaNamespace`, which namespaces the command appropriately.

PR #2644, opened 2017-08-16
Add support for --dev-logger-port in 'ionic serve'

nitrotm:


janpio:
What exactly is a "development logger port"?

nitrotm:
See:

https://github.com/ionic-team/ionic-app-scripts/blob/master/src/serve.ts
https://github.com/ionic-team/ionic-app-scripts/blob/master/src/dev-server/notification-server.ts

It's also called 'notification server'. When running ionic serve, there are 2 websockets ports. One is used for the live-reload, the other (notification server) seems to be used to notify the browser of server side messages, such as ionic app compilation issues.

But I don't know exactly to be honest as I haven't design nor implemented this feature.

This pull request fix the missing option in ionic-cli to specify the dev-logger-port. Something which is necessary when running/serving more than 1 app concurrently during development.

janpio:
Ok, interesting. 
Did this feature get lost from 3.7.0 to 3.8.x?
Do you manually have to set it?

nitrotm:
I guess it was lost due to some refactoring, because since I upgraded to 3.8.x, the option is ignored by ionic-cli and so only the default dev-logger-port is used.
My setup is using docker containers to run each app separately. I then expose the 3 ports for each (ie. 8100, 35729, 53703) on my host were I code. I can remap the port numbers but ionic 'encode' the ports in a <script> tag in index.html. So what happen is that I get the logs and errors from the other apps and that mess up everything.
It could be that if you run two apps on the same machine without containers, the script detects the port is used and change it dynamically. But that can't function with containers.

dwieeb:
Thanks @nitrotm. I'm not entirely familiar with the internals of app-scripts myself. A lot of the app-scripts options are now translated from the CLI options here: https://github.com/ionic-team/ionic-cli/blob/master/packages/@ionic/cli-utils/src/lib/ionic-angular/serve.ts#L80-L102 as you discovered.

mcanfield:
@nitrotm Thank you for doing the legwork on this. I just updated my Docker images this morning and development flow was broken due to this regression.

@dwieeb Thanks for merging, this was indeed a regression. It definitely worked before. It's a value that needs to be changed anytime multiple `ionic serve` instances are running. The scenario @nitrotm outlines with Docker containers is status quo for my team. 

What's the timeline on `3.10.0`? I'm not seeing a due date on the milestone and it says 100% complete.

dwieeb:
@mcanfield Hoping for today, actually.

PR #2692, opened 2017-08-23
fix cordova "build after hook"

rtwall:
`cordova prepare` is called without the `platform` argument, breaking
platform specific builds.


dwieeb:
How does it break platform-specific builds?

rtwall:
@dwieeb 

`cordova prepare` was failing on alpine linux as it was attempting to prepare`ios` even though I ran `ionic cordova build android`

```
Updating CocoaPods specs repo
Failed to install 'cordova-plugin-intercom': undefined
CocoaPods was not found. Please install version 1.0.1 or greater from https://cocoapods.org/
```

I just tried again and it's now continuing past the failure.  Probably shouldn't be attempting `ios` since I specified `android`?

dwieeb:
Gotcha--thanks. I will merge this!

PR #2718, opened 2017-09-07
feat(serve): adds devapp support

manucorporat:


package.json
```diff
@@ -27,5 +27,8 @@
     "tslint": "^5.7.0",
     "tslint-ionic-rules": "^0.0.11",
     "typescript": "^2.5.2"
+  },
+  "dependencies": {
+    "@ionic/discover": "^0.2.1"
```
dwieeb: This has to go in `packages/ionic/package.json`.

packages/ionic/src/commands/serve.ts
```diff
@@ -3,6 +3,7 @@ import * as chalk from 'chalk';
 import { CommandLineInputs, CommandLineOptions } from '@ionic/cli-utils';
 import { Command, CommandMetadata } from '@ionic/cli-utils/lib/command';
 import { BIND_ALL_ADDRESS, BROWSERS, DEFAULT_DEV_LOGGER_PORT, DEFAULT_LIVERELOAD_PORT, DEFAULT_SERVER_PORT } from '@ionic/cli-utils/lib/serve';
+import { Publisher } from '@ionic/discover';
```
dwieeb: Please use a dynamic import in `startDevApp` to only import when `Publisher` is required.

```javascript
const { Publisher } = await import('@ionic/discover');
```

dwieeb:
Merged!

PR #2743, opened 2017-09-16
chore(resources): remove outdated comment

janpio:
outcommented variable is not present anywhere, comment is also wrong, so no need to keep it around


dwieeb:
Thanks!

PR #2914, opened 2017-12-10
Update start.ts

Tommertom:
Add the preview app which is an excellent showcase of all the beautifull UI components Ionic Framework has to offer


dwieeb:
Hi @Tommertom! Thanks for your interest in adding the preview app to `ionic start`. For the next version of the CLI (which will likely be 4.0), we're advocating that people use repository URLs for fully functional apps (as opposed to starter templates) such as the preview app and the conference app.

The command, for example, will be `ionic start <name> https://github.com/ionic-team/ionic-preview-app`. (this already works in the latest CLI).

I'm going to close this, but I'll add a reminder to myself to add the preview app to the documentation alongside the conference app so more people can find it and know about it. Thanks! 👍

Issue #2967, opened 2018-01-25
Contributing from Windows - npm run bootstrap - 'rm' is not recognized as an internal or external command


Tiberriver256:
Attempting to run the following command from the [contributing](https://github.com/ionic-team/ionic-cli/blob/master/CONTRIBUTING.md) guidelines:

```js
npm run bootstrap
```
I received the following error:

```
'rm' is not recognized as an internal or external command
```

**Full Error**
```
> @ bootstrap C:\My\ionic-cli
> lerna bootstrap && npm run build

lerna info version 2.8.0
lerna info versioning independent
lerna info Bootstrapping 9 packages
lerna info lifecycle preinstall
lerna info Installing external dependencies
lerna info Symlinking packages and binaries
lerna info lifecycle postinstall
lerna info lifecycle prepublish
lerna info lifecycle prepare
lerna success Bootstrapped 9 packages

> @ build C:\My\ionic-cli
> lerna run build

lerna info version 2.8.0
lerna info versioning independent
lerna ERR! build Errored while running script in '@ionic/cli-framework'
lerna ERR! execute callback with error
lerna ERR! Error: Command failed: npm run build
lerna ERR! 'rm' is not recognized as an internal or external command,
lerna ERR! operable program or batch file.
lerna ERR! npm ERR! code ELIFECYCLE
lerna ERR! npm ERR! errno 1
lerna ERR! npm ERR! @ionic/cli-framework@0.1.2 clean: `rm -rf index.* definitions.* guards.* ./lib ./utils`
lerna ERR! npm ERR! Exit status 1
lerna ERR! npm ERR!
lerna ERR! npm ERR! Failed at the @ionic/cli-framework@0.1.2 clean script.
lerna ERR! npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
lerna ERR!
lerna ERR! npm ERR! A complete log of this run can be found in:
lerna ERR! npm ERR!     C:\Users\WMRM\AppData\Roaming\npm-cache\_logs\2018-01-25T01_07_35_648Z-debug.log
lerna ERR! npm ERR! code ELIFECYCLE
lerna ERR! npm ERR! errno 1
lerna ERR! npm ERR! @ionic/cli-framework@0.1.2 build: `npm run clean && tsc`
lerna ERR! npm ERR! Exit status 1
lerna ERR! npm ERR!
lerna ERR! npm ERR! Failed at the @ionic/cli-framework@0.1.2 build script.
[...]

PR #2968, opened 2018-01-25
Minor changes to make npm commands run on Windows -- Issue #2967

Tiberriver256:
Fix is for Issue #2967 

-- Replaced all 'rm -rf' with rimraf
-- Replaced single quote in lerna exec npm script with double quote to make Windows happy

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,22 +5,16 @@
     "clean": "lerna run clean",
     "lint": "lerna run lint",
     "lint:fix": "lerna run lint -- -- --fix",
-    "link":
-      "lerna exec --scope=ionic --scope=\"@ionic/+(lab|v1-util)\" npm link",
-    "unlink":
-      "lerna exec --scope=ionic --scope=\"@ionic/+(lab|v1-util)\" npm unlink",
+    "link": "lerna exec --scope=ionic --scope=\"@ionic/+(lab|v1-util)\" npm link",
+    "unlink": "lerna exec --scope=ionic --scope=\"@ionic/+(lab|v1-util)\" npm unlink",
     "test": "lerna run test",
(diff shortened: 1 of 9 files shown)


dwieeb:
Sorry for the giant red X. I don't really like that UX choice by github.

Tiberriver256:
No worries. 

I was reading a bit on the Lerna, am I understanding that right that if I install it as a dev dependency in the root package lerna will make sure it's installed in all the sub-packages?

dwieeb:
It doesn't install it in the subpackages. We're actually not doing what they recommend (read https://github.com/lerna/lerna#common-devdependencies). Lerna recommends doing what you've done with `rimraf`, but for now, we're running npm scripts _only_ in the root package, which alters PATH and lets you use `rimraf`, `tsc`, `tslint`, `jest`, etc when using `lerna run`.

For example if the `npm run clean` command is run in the root, it adds `<root>/node_modules/.bin` (where all the binaries are) to PATH, runs `lerna run clean`, which runs each package's "clean" npm script, which references the `rimraf` binary. So it works as long as you don't run `npm run clean` in the subpackages. 

What we're doing is not ideal. What you've done is probably better, but we'll have to add all packages with binaries to each subpackage to be consistent. I'll merge it. 👍

dwieeb:
@Tiberriver256 I found a BOM inserted in package.json. Please configure your text editor to no include BOMs as they can have super weird effects on files, at least in Linux/BSD systems: https://github.com/ionic-team/ionic-cli/commit/e2e1465f890aac73fe119cef3788e8d1b25dd617

more info: https://en.wikipedia.org/wiki/Byte_order_mark

Issue #3019, opened 2018-03-20
Help Wanted: Test CLI tooling for Ionic Angular 4


dwieeb:
Hi all! 👋 

We'd love to have some help testing **the tooling** for Ionic Angular 4. This issue will document how to get started.

#### Start Testing

Make sure you have the latest CLI release candidate:

```
npm i -g ionic@rc
```

Enable the `angular` project type:

```
ionic config set -g features.project-angular true
```

Start a new app, choosing `angular` when prompted for project type:

```
ionic start
```

We don't yet have a migration guide for existing apps, but we are keeping [this list of breaking changes](https://github.com/ionic-team/ionic/blob/master/angular/BREAKING.md) up to date.

#### Found an Issue?

Feel free to reply to this thread. If it's complicated and you need more room, feel free to open a new issue.

If you find an issue with the framework, please open an issue in [the framework repo](https://github.com/ionic-team/ionic).

rdlabo:
It's great! I could create v4 project and serve!!
But I found the function `ionic generate` is dead. so  I will report it.

```
$ ionic g
[WARN] Could not load schematics for ionic generate. Use --verbose to debug.
[ERROR] No suitable schematics found.
```

```
$ ionic g --verbose
  ionic:cli-utils:lib:project Project type from config: Ionic Angular v4+ (angular) +0ms
  ionic:cli-utils CLI flags: { interactive: true, confirm: false } +0ms
  ionic:cli { binPath: '/usr/local/lib/node_modules/ionic/bin/ionic', libPath: '/usr/local/lib/node_modules/ionic/dist/index.js', execPath: '/Users/sakakibara/devTry/ionic4-rc0', version: '4.0.0-rc.0' } +0ms
[WARN] Could not load schematics for ionic generate. Use --verbose to debug.
  ionic:cli-utils:lib:project:angular:generate Error while loading schematics: Error: Cannot find module '@angular-devkit/schematics'
  ionic:cli-utils:lib:project:angular:generate     at Function.Module._resolveFilename (module.js:536:15)
  ionic:cli-utils:lib:project:angular:generate     at Function.Module._load (module.js:466:25)
  ionic:cli-utils:lib:project:angular:generate     at Module.require (module.js:579:17)
  ionic:cli-utils:lib:project:angular:generate     at require (internal/module.js:11:18)
[...]

PR #3023, opened 2018-03-23
import local @angular-devkit/schematics

rdlabo:
Default import @angular-devkit/schematics use global one.

https://github.com/ionic-team/starters/commit/155d3f6ce4f8c8a95317df683409050d06457d00 commit is locally. so change import directory.

https://github.com/ionic-team/ionic-cli/issues/3019#issuecomment-375528179

### DIFF
--- a/packages/@ionic/cli-utils/src/lib/project/angular/generate.ts
+++ b/packages/@ionic/cli-utils/src/lib/project/angular/generate.ts
@@ -208,8 +208,8 @@ ${chalk.cyan('[1]')}: ${chalk.bold('https://ionicframework.com/docs/cli/projects
   private async getSchematics(): Promise<Schematic[]> {
     if (!this.schematics) {
       try {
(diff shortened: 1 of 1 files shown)


dwieeb:
Thanks for taking a look and finding the issue, @rdlabo!! 

Would you mind changing it slightly to work like this example:
* https://github.com/ionic-team/ionic-cli/blob/master/packages/%40ionic/cli-utils/src/lib/project/ionic-angular/generate.ts#L94
* https://github.com/ionic-team/ionic-cli/blob/master/packages/%40ionic/cli-utils/src/lib/project/ionic-angular/app-scripts.ts#L6-L10

Doing this will make sure it 1) uses node modules resolution algorithm and 2) work for cases when the dev is in a subfolder of the project.

Does that sound reasonable? 😄

rdlabo:
thanks for review and reply.

I made new pull request https://github.com/ionic-team/ionic-cli/pull/3024 and this commit is close.

PR #3024, opened 2018-03-23
import local @angular-devkit/schematics

rdlabo:
This commit is update by https://github.com/ionic-team/ionic-cli/pull/3023/commits/8a83c6b844fae177be6fe221a6acdb96643af87d .

use Function importAppScripts.

### DIFF
--- a/packages/@ionic/cli-utils/src/lib/project/angular/app-scripts.ts
+++ b/packages/@ionic/cli-utils/src/lib/project/angular/app-scripts.ts
@@ -1,0 +1,8 @@
+import { compileNodeModulesPaths, resolve } from '@ionic/cli-framework/utils/npm';
+
+export async function importAppScripts(projectDir: string): Promise<any> {
+  const appScriptsPath = resolve('@ionic/app-scripts', { paths: compileNodeModulesPaths(projectDir) });
+
+  return require(appScriptsPath);
+}
+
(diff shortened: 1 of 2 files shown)


packages/@ionic/cli-utils/src/lib/project/angular/app-scripts.ts
```diff
@@ -0,0 +1,8 @@
+import { compileNodeModulesPaths, resolve } from '@ionic/cli-framework/utils/npm';
+
+export async function importAppScripts(projectDir: string): Promise<any> {
```
dwieeb: Define two functions, one for importing `@angular-devkit/schematics` and one for `@angular-devkit/schematics/tools`.

packages/@ionic/cli-utils/src/lib/project/angular/generate.ts
```diff
@@ -208,8 +209,8 @@ ${chalk.cyan('[1]')}: ${chalk.bold('https://ionicframework.com/docs/cli/projects
   private async getSchematics(): Promise<Schematic[]> {
     if (!this.schematics) {
       try {
-        const { SchematicEngine } = await import('@angular-devkit/schematics');
-        const { NodeModulesEngineHost } = await import('@angular-devkit/schematics/tools');
+        const { SchematicEngine } = await importAppScripts('@angular-devkit/schematics');
```
dwieeb: You'll need to pass in `this.project.directory` to the two functions, not the module name.

dwieeb:
Sorry, with my comment I was trying to give you an example of a function that has one parameter, the project directory (local), and then uses that to configure `resolve` to look for the module in the local `node_modules`.

`app-scripts.ts` and `importAppScripts` refers to `@ionic/app-scripts`, but we'll be importing two modules of the `@angular-devkit/schematics` package.

PR #3028, opened 2018-03-24
change dev-env

rdlabo:
1. add how to deal with errors
2. ignore local env file.


CONTRIBUTING.md
```diff
@@ -79,6 +79,7 @@ Each `packages/*` folder represents a package on npm.
 1. TypeScript source files are in `packages/**/src`.
 1. Good luck! :muscle: Please open an issue if you have questions or something
    is unclear.
+1. If node-sass error occurs, you should add permission `npm`. ex) `sudo chown -R $(whoami) ~/.npm`
```
dwieeb: Hmm... the Ionic CLI doesn't use `node-sass`. Are you sure this wasn't an issue from something else?

packages/@ionic/lab/.gitignore
```diff
@@ -1 +1,2 @@
 www
+bin/ionic-lab.*
```
dwieeb: That's strange, which file(s) are generated?

CONTRIBUTING.md
rdlabo: Dependencies @stencil/core. It was installed `node-sass` with dependency resolution.
Maybe @dwieeb 's `@ionic/lab/package-lock.json` written.

packages/@ionic/lab/.gitignore
rdlabo: generate `ionic-lab.3077068012`. The contents are the same as `ionic-lab`.

dwieeb: Did it do this with any of the other packages with binaries? `v1-toolkit`, `cli-scripts`? Just curious.

CONTRIBUTING.md
dwieeb: Oops... it must be Stencil. I forgot I used SCSS for Ionic Lab 😆

packages/@ionic/lab/.gitignore
rdlabo: I retry new project same operation, but don't make this file.
Probably, If node-sass error occurs, this file made.

CONTRIBUTING.md
rdlabo: It is the package causes builds to fail often in another project 😢

dwieeb: @rdlabo I'm worried this is a bit too specific to what happened in your set up. Perhaps we can link to common installation issues with node-sass (https://github.com/sass/node-sass/blob/master/TROUBLESHOOTING.md#installation-problems) ?

Or maybe leave it out because hopefully soon I'll have time to switch Ionic Lab to just use CSS, then it wouldn't need `node-sass`.

What do you think?

rdlabo: the problems happened with  `$sudo npm run link` . and I don't found installation issues with node-sass this issue. (maybe happened because passed through lerna)

mumu... In the first place, using `$sudo npm` normally is security is not good. 
So I think that it is kindness to write how to add permission `npm` in the guide.
(Apart from that, node_sass frequently gives warnings for each version, so welcome to go out of the package!)

PR #3027, opened 2018-03-23
import local @angular-devkit/schematics

rdlabo:
Change define two functions.

https://github.com/ionic-team/ionic-cli/pull/3024/files/34466b25a0ef8b5802495de2fe96f00406897d85#diff-3046d014b39b25dfe813bf43052acb19

I'm sorry I could only do source reading because I can't build it. Thanks for instruct.

PR #3030, opened 2018-03-24
fix(generate): use local @angular-devkit/schematics/tools

rdlabo:
This is https://github.com/ionic-team/ionic-cli/pull/3027 's bug fix.

resolve( https://github.com/ionic-team/ionic-cli/blob/master/packages/%40ionic/cli-framework/src/utils/npm.ts#L72 ) needs `package.json` in case path is folder.
So specify index.js.


dwieeb:
Thanks @rdlabo, I haven't had a chance to test this. I think the `resolve` function needs some updates, but I'll look at it later.

dwieeb:
@rdlabo I refactored the function so it works without "index.js": https://github.com/ionic-team/ionic-cli/commit/b803818971022e91a0020e57b9732eee1600f9a9

It works for me! Please test.

rdlabo:
thanks. I gave up because I did not know the scope of the impact, so I thank you for refactoring!

and It works well for me!

PR #3031, opened 2018-03-26
change generate template

rdlabo:
1. selector match https://github.com/ionic-team/starters/blob/master/angular/official/blank/src/app/pages/home/home.page.ts 

fix https://github.com/ionic-team/ionic-cli/issues/3019#issuecomment-375599108

2. default routing NavController is include

### DIFF
--- a/packages/@ionic/schematics-angular/page/files/__path__/__name@kebabCase@if-flat__/__name@kebabCase__.page.ts
+++ b/packages/@ionic/schematics-angular/page/files/__path__/__name@kebabCase@if-flat__/__name@kebabCase__.page.ts
@@ -1,7 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 
 @Component({
-    selector: '<%= kebabCase(name) %>-page',
+    selector: 'app-page-<%= kebabCase(name) %>',
     templateUrl: './<%= kebabCase(name) %>.page.html',
     styleUrls: ['./<%= kebabCase(name) %>.page.<%= styleext %>'],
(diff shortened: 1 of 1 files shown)


rdlabo:
Ionic v3 way is generate 1 page 1 module. Perhaps we should include module file in the template...

I think this is a problem directly linked to Ionic v4 router. Does Ionic team have a policy?

rdlabo:
@ionic/angular@0.1.4 don't have `NavController` . This don't work now.

dwieeb:
> @0.1.4 don't have NavController . This don't work now.

Don't worry about it. We will have to update the starter and these generators anyway with the framework release I'm expecting today.

dwieeb:
@rdlabo Actually, if you do remove the `NavController` stuff, I'll merge this in right away.

rdlabo:
@dwieeb Sure. @0.1.4 seems change around router, so I will propose `NavController` on n another occasion.
I changed this pull request now.

rdlabo:
@dwieeb If scheduled to release soon, I'm glad you check this request: https://github.com/ionic-team/ionic/pull/14217 
It is important issue for Japanese users  (and perhaps, using 2 byte character world users).

dwieeb:
@rdlabo Thanks!

I will bring up the negative margin issue. It seems like a big deal!

rdlabo:
@dwieeb thanks very much!! in  I giving up this because great deal of change in v3, so I'm glad to receive the examination!

PR #3033, opened 2018-03-26
when generate page, add page entryComponents

rdlabo:
When using NavController in no LazyLoading, target Page must be registered ngModule entryComponents.
I add function `ionic generate page` add ngModule entryComponents.


dwieeb:
@rdlabo I agree, we should be doing this for pages that are not lazy loaded. 

Perhaps in the future we can have an option to generate a page that _is_ lazy loaded.

Thank you!

PR #3034, opened 2018-03-26
WIP: add new Function generate `ion-component`

rdlabo:
`ng generate component` make folder under `src/app`. It's useful when advanced users create complicated applications. But I like Ionic way make forlder under `src/components/`. This was very easy to use for beginners and easy to use.

So I propose CLI generate component function look like v3 CLI.

use `ionic g ion-componet`

1) This request works but is incomplete
2) I wanted to automatically create `ion-components/ion-components.module.ts`, but I can't make it...


rdlabo:
@dwieeb `ionic generate component` in v3 is locked folder `src/components` and use `components.module.ts` . I think this is very easy for learning.

But `ionic generate component` in v4 ( = `ng generate component` ) can  be placed anywhere, and use `app.module.ts`.

I think this is difficult for beginner.  so want to command similar to v3.

dwieeb:
@rdlabo Ah, yes. We can use the `--module` option for the `component` schematic: https://github.com/angular/devkit/blob/master/packages/schematics/angular/component/index.ts#L40

rdlabo:
@dwieeb Thanks reply. I changed `ionic g ion-component` make folder `src/app/ion-components` `src/app/components` . (It matched the behavior of CLI v3)

> We can use the --module option for the component schematic: 

This option designation of an already existing module, it could not be created.

ex)
```
$ionic generate component example --module=components/components.module.ts
Error: Specified module does not exist
Specified module does not exist
[ERROR] Could not generate component.
```

rdlabo:
@dwieeb 

> I don't like having two folders for the same type of component.

I changed factory component `ion-components` to `components`. now, `ion-component` command make files under `app/components/` folder.
(Does my understanding have it? )

>  then --module will work.

This is good idea! I want you to do so.
( this command is `ionic start --module` , is this it?

rdlabo:
add my repository https://github.com/rdlabo/ionic-cli/tree/add_module_flg .
I'll change to Implementation.

dwieeb:
@rdlabo I think we're having some miscommunication again. Let me try to make some progress on this and then we'll go from there.

rdlabo:
@dwieeb Oh...thanks for pointing out. I will wait for it.

dwieeb:
@rdlabo I think I was misunderstanding the original issue. 

You're basically saying that having app components in `src/app/` (vs. just `src/`, like in v3) is confusing for beginners. Can you explain why that is? `ng new` creates a similarly structured project, as well. 

When you get a chance, please test out some recent changes in the new starter. 

1. Pull down `master` and build, link
1. `ionic start` the new angular starter
1. Don't forget to `npm link @ionic/schematics/angular`

rdlabo:
I tried recent `ionic g component` and this is great!!
before, `$ionic g component test` made `app/test`. I expected `app/components/test`. but now! it is realized!!

this problem was solved, and I close this. thanks!!

PR #3037, opened 2018-03-26
WIP: Capacitor support

mlynch:
Work in progress PR for adding Capacitor support into the Ionic CLI (under the beta flag)

Don't merge yet, but feedback welcome


packages/ionic/src/commands/capacitor/base.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  async preRunChecks(runinfo: CommandInstanceInfo) {
+    runinfo;
+  }
+
+  async runCapacitor(argList: string[], { fatalOnNotFound = false, truncateErrorOutput = 5000, ...options }: IShellRunOptions = {}): Promise<void> {
+    const { ERROR_SHELL_COMMAND_NOT_FOUND } = await import('@ionic/cli-utils/lib/shell');
+    try {
+      this.env.close();
+      await this.env.shell.run('npx', ['capacitor'].concat(argList), { fatalOnNotFound, truncateErrorOutput, stdio: 'inherit', ...options });
```
dwieeb: You shouldn't need to use `npx`, if you prefer. The `.run()` utility adds `node_modules/.bin` to `PATH` in the same way npx does.

packages/ionic/src/commands/capacitor/base.ts
```diff
@@ -0,0 +1,47 @@
...
+      this.env.open();
+    } catch (e) {
+      this.env.open();
+      if (e === ERROR_SHELL_COMMAND_NOT_FOUND) {
+        throw new FatalException(
+          `Unable to run npx, the new command execution system in NPM.
+          This means your Node version is too old. Capacitor requires Node 8.6.0 or greater,
```
dwieeb: Is this node requirement because of the Capacitor CLI codebase, or `npx`?

packages/@ionic/cli-utils/src/definitions.ts
```diff
@@ -143,8 +143,12 @@ export interface ProjectCordovaIntegration extends ProjectIntegration {
   setupEngineHooks?: boolean;
 }
 
+export interface ProjectCapacitorIntegration extends ProjectIntegration {
+}
+
 export interface ProjectIntegrations {
   cordova?: ProjectCordovaIntegration;
+  capacitor?: ProjectCapacitorIntegration;
```
dwieeb: You can put integration files here: https://github.com/ionic-team/starters/tree/master/integrations

Then you'll get a url for the `.tar.gz`, which you can use in these files: https://github.com/ionic-team/ionic-cli/tree/master/packages/%40ionic/cli-utils/src/lib/integrations

packages/ionic/src/commands/capacitor/base.ts
mlynch: Cool should be more stable to just call the binary directly

packages/ionic/src/commands/capacitor/base.ts
mlynch: Without npx as above it won't be a requirement (but yes, somewhere in Node 5 npx was added as a command without having to install something else)

janpio:
How about `ionic cap` aliases for all these? (Being used to `npx cap` and all...)

dwieeb:
@mlynch If you pull in `master` and then add this alias mapping to the `NamespaceMap` here: https://github.com/ionic-team/ionic-cli/blob/master/packages/ionic/src/commands/index.ts#L12-L21 then `ionic cap` will be an alias

```ts
['cap', 'capacitor'],
```

dwieeb:
Merged in. Enable the beta capacitor commands with:

```
ionic config set -g features.capacitor-commands true
```

PR #3045, opened 2018-03-28
refactor(cli-utils) get npm package method

rdlabo:
gathered duplicate methods into one. (Corresponds to package name change)


packages/@ionic/cli-utils/src/lib/project/angular/index.ts
```diff
@@ -89,75 +89,26 @@ export class Project extends BaseProject {
     return registry;
   }
 
-  async getFrameworkVersion() {
-    const pkgName = '@ionic/angular';
+  async getPackageVersion(pkgName: 'ionicAngular' | 'ionicCore' | 'ionicSchematicsAngular' |
```
dwieeb: This class is one of three project classes:
- `src/lib/project/angular/index.ts`
- `src/lib/project/ionic-angular/index.ts`
- `src/lib/project/ionic1/index.ts`

They all extend `BaseProject`, which is in `src/lib/project/index.ts`, and they all have methods like this. Maybe now is a good time to refactor them to use a single utility method?

Also you don't need to use `'ionicAngular' | 'ionicCore' ...` as the string literal, you can just use the package name itself: `'@ionic/angular' | '@ionic/core' ...` 😄

rdlabo:
`BaseProject in src/lib/project/index.ts` use `getInfo()` (in three project) only.
So I think not need to make it common name.

If `getFrameworkVersion()` seems necessary in the future , I will change this refactor leave `getFrameworkVersion()` .

> you can just use the package name itself

thanks. I will do so!

rdlabo:
@dwieeb  when you're available, please indicate whether to leave `getFrameworkVersion()` .

dwieeb:
@rdlabo Sorry--I think there was a misunderstanding. 

I am trying to say that `getPackageVersion()` can be moved into `BaseProject` and used in all the subclasses to get the versions of a package.

It doesn't need to be done _now_, though. What you have is great. If you change those string literals I'll merge it!

rdlabo:
Oh, I understand.  sorry for English reading comprehension is low...

I changed. and this line work changed.
https://github.com/ionic-team/ionic-cli/pull/3045/commits/747e66fd2c4e89ebef2bc6d9394dea33751df057#diff-49365df12c59dc647697ca9f0e2ff4b5L36

rdlabo:
@dwieeb I also made common methods. But this method is leave.

https://github.com/ionic-team/ionic-cli/blob/master/packages/%40ionic/cli-utils/src/lib/project/ionic1/index.ts#L84

dwieeb:
Awesome work @rdlabo !! 🎉 

> It changed. and this line work changed. (maybe mistake)
747e66f#diff-49365df12c59dc647697ca9f0e2ff4b5L36

Yes, this was a typo on my part. Thanks for fixing it!

PR #3057, opened 2018-04-03
WIP: feat(cli-utils) remove @ionic/core from cli version check.

rdlabo:
> @ionic/core won't be in the list of dependencies for starters in the stable release
I left it as pull request, so as not to forget it.


dwieeb:
Ah yeah, we ultimately won't need this. Thanks!

PR #3159, opened 2018-05-11
feat(@ionic/cli): Support multiple Angular projects and configurations in a single workspace

stupidawesome:
This PR adds support for specifying the Angular project/configuration option when serving or building through Ionic.

Angular CLI documentation:
https://github.com/angular/angular-cli/wiki/serve
https://github.com/angular/angular-cli/wiki/build

`target` and `environment` has been superseded in v6 and replaced with `configuration`.

The `angular` project type currently errors when trying to use a project name other than `app`. This can now be configured using the `defaultProject` setting in `angular.json`. If there are multiple angular apps, the target angular project can now be specified using the `project` option.

Examples:
```
# New options
ionic serve --project app --configuration staging
ionic build --project app -c staging
ionic cordova build/run --project app -c staging

# Already supported
ionic generate --project app
```


packages/@ionic/cli-utils/src/lib/project/angular/utils.ts
```diff
@@ -24,19 +24,19 @@ export async function readAngularConfigFile(p: string): Promise<AngularConfig> {
...
 
-export async function addCordovaEngineForAngular(project: IProject, platform: string): Promise<void> {
+export async function addCordovaEngineForAngular(project: IProject, platform: string, appName = 'app'): Promise<void> {
   debug('Adding Cordova engine for platform: %s', platform);
-  const srcDir = await project.getSourceDir();
   const platformWWW = path.resolve(project.directory, 'platforms', platform, 'platform_www');
   const angularJsonPath = path.resolve(project.directory, ANGULAR_CONFIG_FILE);
   const angularJson = await readAngularConfigFile(angularJsonPath);
-  const app = angularJson.projects.app;
+  const angularApp = angularJson.projects[angularJson.defaultProject || appName];
```
dwieeb: Shouldn't we use `appName` if provided, otherwise use `angularJson.defaultProject`? This seems like it wouldn't work if a `defaultProject` is defined but `--project=something` is used.

packages/ionic/src/commands/cordova/run.ts
```diff
@@ -93,6 +94,8 @@ export class RunCommand extends CordovaCommand implements CommandPreRun {
         groups: [OptionGroup.Advanced, OptionGroup.Cordova],
         hint: 'cordova',
       },
+      // Angular build options
+      ...ANGULAR_BUILD_COMMAND_OPTIONS,
```
dwieeb: These additional options should be returned in the `getCommandMetadata` method of `ServeRunner` for the angular project type so they don't show up for non-angular projects.

packages/ionic/src/commands/cordova/prepare.ts
```diff
@@ -37,6 +38,7 @@ You may wish to use ${chalk.green('ionic cordova prepare')} if you run your proj
           default: true,
         },
         ...APP_SCRIPTS_OPTIONS,
+        ...ANGULAR_BUILD_COMMAND_OPTIONS,
```
dwieeb: This also shouldn't be here, exactly, but it's my fault. I must've missed the `APP_SCRIPTS_OPTIONS` for cordova prepare. So, it's fine for now.

stupidawesome:
@dwieeb Thankyou for your feedback. I've amended the those sections and a few more places I had missed. While I was working on this I realised I could just pass through all of the Angular CLI options so I went ahead and did that. Hopefully this can be moved into the ng-toolkit later on.

dwieeb:
@stupidawesome I was hiding the plethora of build and serve options for a couple reasons.
- By not including them, we don't have the maintenance overhead of keeping our mapping perfectly up to date.
- I considered them advanced options. Having them listed is a bit overwhelming for new users and probably not useful for a huge majority. I imagine the output of `ionic serve --help` is pretty big now! 😆 I had a sentence about this in the serve help:

    > Options not listed below are considered advanced and can be passed to the Angular CLI using the '--' separator after the Ionic CLI arguments.

I am also trying to keep this code relatively minimal and stagnant so that it's easier to move into the upcoming ng-toolkit and then expand upon it.

You must be having fun working on the CLI. 😄 This is a huge amount of work. I'll review it in depth tomorrow. If you want to help with the ng-toolkit, I'd love to chat! If you're on our Ionic Worldwide slack, I'm dan.

stupidawesome:
@dwieeb Yeah, I figure most of this will carry over into the ng-toolkit. In that regard I'd rather have the CLI be explicit about which options are supported, even if it's a little verbose. I need these features for the apps I'm working on so I'll catch you on slack.

stupidawesome:
Thanks @dwieeb, that makes sense now. Please see the amended PR.

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
```diff
@@ -1,15 +1,41 @@
...
 
 const debug = Debug('ionic:cli-utils:lib:project:angular:build');
 
+// tslint:disable no-null-keyword
+export const NG_BUILD_OPTIONS = [
+  {
+    name: 'prod',
+    summary: `Flag to set configuration to ${chalk.green('prod')}`,
+    type: Boolean,
+    default: null,
```
dwieeb: Reason for `default: null` ?

packages/@ionic/cli-utils/src/definitions.ts
```diff
@@ -546,8 +546,44 @@ export interface BuildOptions<T extends ProjectType> extends BaseBuildOptions {
 }
 
 export interface AngularBuildOptions extends BuildOptions<'angular'> {
-  target?: string;
-  environment?: string;
+  'prod'?: boolean;
+  'project'?: string;
+  'configuration'?: string;
+  'main'?: string;
```
dwieeb: Missed some types

packages/@ionic/cli-utils/src/definitions.ts
```diff
@@ -601,8 +637,36 @@ export interface ServeOptions {
 }
 
 export interface AngularServeOptions extends ServeOptions {
-  target?: string;
-  environment?: string;
+  'project'?: string;
+  'prod'?: string;
+  'configuration'?: string;
+  'browser-target'?: string;
```
dwieeb: here as well

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
```diff
@@ -18,73 +44,45 @@ export class BuildRunner extends BaseBuildRunner<AngularBuildOptions> {
...
-    const environment = options['environment'] ? String(options['environment']) : undefined;
-
-    if (!target) {
-      if (options['dev']) {
-        target = 'development';
-      } else if (options['prod']) {
-        target = 'production';
-      }
-    }
+    const ngOptions = NG_BUILD_OPTIONS
```
dwieeb: We do lose some type info here by doing this, but that's alright if there's tests, which there are

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
```diff
@@ -18,73 +44,45 @@ export class BuildRunner extends BaseBuildRunner<AngularBuildOptions> {
...
+        accum[option] = String(options[option]);
+
+        return accum;
+      }, {} as any);
+
     const args: ParsedArgs = {
       _: [],
-      environment: options.environment,
+      ...ngOptions,
+      'output-path': 'www',
```
dwieeb: I believe this is specified in angular.json. No reason to specify it here

packages/ionic/src/commands/cordova/run.ts
```diff
@@ -95,16 +124,6 @@ export class RunCommand extends CordovaCommand implements CommandPreRun {
       },
     ];
 
-    try {
```
dwieeb: By doing this, I think you are throwing away the dynamic help output. `--help` uses `getMetadata()`

packages/ionic/src/commands/cordova/run.ts
```diff
@@ -14,20 +15,54 @@ import {
...
 export class RunCommand extends CordovaCommand implements CommandPreRun {
-  protected serveRunner?: ServeRunner<ServeOptions>;
+  protected runner?: ServeRunner<ServeOptions> | BuildRunner<BuildOptions<any>>;
 
-  async getRunner() {
-    if (!this.serveRunner) {
-      this.serveRunner = await ServeRunner.createFromProjectType(this.env, this.env.project.type);
+  async getRunner(livereload: boolean) {
+    if (!this.runner) {
+      this.runner = livereload
```
dwieeb: ah, good catch

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
stupidawesome: I was thinking for a CI/CD scenario it might be useful to be able to run `ng build` separately which could build each app to a separate folder (as configured in angular.json), whereas running `ionic cordova build` should always build to `www` so cordova can pick it up.

The alternative here is to require the output path for every app in angular.json to point at `www`, or copy folders from the output path into `www` during the build process.

dwieeb: Ohh, okay. Please check if `options.engine === 'cordova'` then. This logic is reused for `ionic build` as well as `ionic cordova build`

packages/ionic/src/commands/cordova/run.ts
stupidawesome: You're right, that does happen. The problem I was trying to fix here was to get the right metadata based on whether `livereload` is enabled. It passes through to `ng serve` when `livereload` is enabled, and `ng build` when it's not. This is a problem if the build and serve options are different since the CLI filters unknown options. Any ideas?

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
stupidawesome: I'm not sure if I grasped the way the CLI works here; this is what I was trying to achieve:

Say for example I had some additional metadata in `NG_BUILD_OPTIONS`
```
  {
    name: 'build-optimizer',
    summary: `Enable or disable the build optimizer`,
    type: Boolean,
    hint: 'ng',
  },
```
Then if I try to run
```
ionic build --prod
```
This is what the Angular CLI would receive
```
ng build --prod=true --build-optimizer=false
```
Which is bad, because `prod` is a meta flag that sets `build-optimizer=true`. I can't just set the default for `build-optimizer` to `true`, because running `ionic build` on it's own shouldn't enable the build optimizer.

So what I changed here, is line 76 in `@ionic/cli-framework/src/lib/options.ts`
```
    default: option.default !== undefined ? option.default : typeDefaults.get(type),
```
This stops the CLI using default values for options that are falsy. I then set the default value for the boolean options to `null` so I can control how they get passed through in `buildOptionsToNgArgs()`.

The end result is that the Ionic CLI won't pass through boolean flags to the Angular CLI if they have a default value of `null`.

So because of these changes and the way I changed `buildOptionsToNgArgs()` to filter args, it's necessary to set a default of `null` for the boolean options so that they don't default to `false`.

packages/ionic/src/commands/cordova/run.ts
dwieeb: Yeah, there's not really a good solution here. Perhaps leave it the way you have it for now. I'm not too worried

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
dwieeb: Yeah, I think the issue is the filter/reduce that includes the flags with whatever value they have.

Boolean flags in the CLI default to `false` (meaning they were not specified on the command line). If they default to `true`, the CLI does some fancy work (mostly using behavior from minimist) to document the negation of the flag in help output. For example, with a boolean option `--foo` with `default: true`, the help output actually documents the negation: `--no-foo`. If `--no-foo` is specified on the command line, the `--foo` option is `false`, otherwise the value is taken from the default, true.

These changes would introduce something new: the differentiation between a boolean flag that is not set on the command line and one that defaults to `false`. It's probably worth doing because then command options would share a default of `null`, I'm just not certain what the implications of this change are without testing it. So, it's a little scary. The filter/reduce in is cool, but we lose type info and it's less explicit. I think you mostly had the filter/reduce because of the number of options, but now there aren't so many anymore.

stupidawesome: Yeah, I can take this out for now since those options are gone. It won't be a problem until it needs to differentiate between boolean options that are/aren't explicitly passed through the command line.

packages/@ionic/cli-utils/src/lib/project/angular/build.ts
```diff
@@ -1,15 +1,40 @@
...
+import { CommandGroup, OptionGroup } from '../../../constants';
 
 import { AngularBuildOptions, CommandLineInputs, CommandLineOptions, CommandMetadata } from '../../../definitions';
-import { CommandGroup, OptionGroup } from '../../../constants';
 import { BUILD_SCRIPT, BuildRunner as BaseBuildRunner } from '../../build';
 import { addCordovaEngineForAngular, removeCordovaEngineForAngular } from './utils';
 
 const debug = Debug('ionic:cli-utils:lib:project:angular:build');
 
+// tslint:disable no-null-keyword
```
dwieeb: These tslint comments should no longer be needed

packages/@ionic/cli-utils/src/lib/project/angular/serve.ts
```diff
@@ -22,6 +22,39 @@ const DEFAULT_PROGRAM = 'ng';
 const NG_SERVE_CONNECTIVITY_TIMEOUT = 20000; // ms
 const NG_AUTODETECTED_PROXY_FILES = ['proxy.conf.json', 'proxy.conf.js', 'proxy.config.json', 'proxy.config.js'];
 
+// tslint:disable no-null-keyword
+const NG_SERVE_OPTIONS = [
+  {
+    name: 'ssl',
+    summary: 'Use HTTPS for the dev server',
+    type: Boolean,
+    default: null,
```
dwieeb: Still default null here, is this correct?

packages/@ionic/cli-utils/src/lib/project/angular/serve.ts
```diff
@@ -22,6 +22,39 @@ const DEFAULT_PROGRAM = 'ng';
...
+  },
+  {
+    name: 'prod',
+    summary: `Flag to set configuration to ${chalk.green('prod')}`,
+    type: Boolean,
+    default: null,
```
dwieeb: here as well

PR #3219, opened 2018-05-26
fix: fix broken tests on windows

stupidawesome:


dwieeb:
Great work! Looks like the takeaways are use `path.resolve` or `path.normalize` expected path results or mock `path` with `path.posix`

PR #3434, opened 2018-07-30
fix(cli-framework/utils): catch connected socket.

rdlabo:
when don't get `sock.on('connect')` Event, `if (sock.connecting === true)` get connected status. 
It worked well in my environment.

Fixes https://github.com/ionic-team/ionic-cli/issues/3410

### DIFF
--- a/packages/@ionic/cli-framework/src/utils/network.ts
+++ b/packages/@ionic/cli-framework/src/utils/network.ts
@@ -84,6 +84,11 @@ export async function isHostConnectable(host: string, port: number, { timeout }:
         resolve(true);
       });
 
+      if (sock.connecting === true) {
+        sock.destroy();
+        resolve(true);
+      }
+
       sock.on('error', err => {
         reject(err);
       });


dwieeb:
But `sock.connecting` is set to `true` whenever `sock.connect()` is called (https://nodejs.org/api/net.html#net_socket_connecting). This is the same as not attempting connections at all, isn't it?

dwieeb:
This can lead to the ERR_CONNECTION_REFUSED error in Chrome because the HTTP server isn't accepting connections. I believe @amanitequeen ran into this when testing: https://github.com/ionic-team/ionic-cli/issues/3410#issuecomment-408772292

rdlabo:
@dwieeb  thanks for review. and as you pointed out, I confirm sock.connecting is set to true whenever error connect. 

I make another pull request. It worked well in my environment(localhost).  Please review this pull request.
https://github.com/ionic-team/ionic-cli/pull/3444

PR #3436, opened 2018-07-30
add(README): add --type=angular flug guide.

rdlabo:


dwieeb:
I think we should link to the beta docs for starting instead: https://beta.ionicframework.com/docs/building/starting

dwieeb:
I made a commit to link to it: https://github.com/ionic-team/ionic-cli#cli-4-is-here-tada

Issue #3432, opened 2018-07-29
ionic serve --ssl does not serve via https


mharrisn:
**Description:**

Neither `ionic serve --ssl` or `ionic serve -s` appear to have any effect. As a result, I cannot serve with https, which restricts capabilities like geolocation in most modern browsers.

According to this commit, it was included: https://github.com/ionic-team/ionic-cli/commit/ed14f58454cb0ed8d84a2b52a566da400eea40ec

**Steps to Reproduce:**

Create a new Ionic 4 project
Run `ionic serve --ssl`

**Output:**

`ng run app:serve --host=0.0.0.0 --port=8101`
appears to be the resulting command

**My `ionic info`:**

```
Ionic:

   ionic (Ionic CLI)          : 4.0.1 (/usr/local/lib/node_modules/ionic)
   Ionic Framework            : @ionic/angular 4.0.0-beta.0
   @angular-devkit/core       : 0.7.0-rc.3
   @angular-devkit/schematics : 0.7.0-rc.3
   @angular/cli               : 6.0.8
   @ionic/ng-toolkit          : 1.0.0
   @ionic/schematics-angular  : 1.0.1

System:

   NodeJS : v9.4.0 (/usr/local/Cellar/node/9.4.0/bin/node)
   npm    : 6.2.0
   OS     : macOS High Sierra
```

**Other Information:**

dwieeb:
Thanks for the issue!

The changelog at time of release is here: https://github.com/ionic-team/ionic-cli/blob/08dba217e3c091b197593fdd6a57cb360b68183a/packages/ionic/CHANGELOG.md

This is already being tracked here: https://github.com/ionic-team/ionic-cli/issues/3305

dwieeb:
Duplicate of https://github.com/ionic-team/ionic-cli/issues/3305

Issue #3305, opened 2018-06-13
HTTPS support for `ionic serve` and Cordova livereload


dwieeb:
The Angular CLI supports HTTPS configuration. Investigation will be needed to ensure it works simply with Cordova livereload re: option passing, etc.

bleuscyther:
GPS wont work outside of safe env. aka: https.

redplane:
Any solutions for this ? I'm having the same issue with `ionic cordova run browser --livereload`

dwieeb:
@redplane That command (and the Cordova browser platform) isn't fully supported. See https://github.com/ionic-team/ionic-cli/issues/3043. The solution is forthcoming and this issue will be updated when it arrives.

redplane:
Will it be for ionic-3 or ionic 4? Im having a project on ionic 3.9.2, a little big that Im afraid of changing

dwieeb:
No plans to add HTTPS to `@ionic/app-scripts` for Ionic 3.

kurtommy:
Updates?

CJTalbot87:
Hi any update on this? I'm struggling to get the Geo-location to work on the Ionic Dev App with HTTP so to be able to serve over HTTPS would be great assuming that would fix the issue.

dwieeb:
CLI 4.4.0 is out with an `--ssl` flag:

```
ionic serve --ssl
```

There are a few gotchas, which I've documented in the original issue description: https://github.com/ionic-team/ionic-cli/issues/3305#issue-332072032

sebast1219:
So now livereload should work in https with --ssl ?

dwieeb:
@sebast1219 `ionic cordova run` with `--ssl` will work, but it seems the iOS wkwebview doesn't accept self-signed certificates. So, it should work with Android. 

The `--ssl` flag is an experimental feature (and is marked as such in `ionic serve --help`), we expect issues.

phal0r:
@dwieeb 
Thanks for this feature, I wanted to try this out and therefor updated to ionic 4.4.0. I can see (and use) `ionic ssl` commands, but no `ionic serve` with `--ssl` flag when performing `ionic serve --help`.

Using `ionic serve --ssl` works so far, but still in http mode, so the ssl flag is not acknowledged. Do I need to update something else?

dwieeb:
[...]

PR #3437, opened 2018-07-30
fix(project): reflect --ssl option.

rdlabo:
`ionic serve --ssl` will make `ng run app:serve --host=0.0.0.0 --port=8100 --ssl`.

Fixes https://github.com/ionic-team/ionic-cli/issues/3432

### DIFF
--- a/packages/@ionic/cli-utils/src/lib/project/angular/serve.ts
+++ b/packages/@ionic/cli-utils/src/lib/project/angular/serve.ts
@@ -99,7 +99,7 @@ ${chalk.cyan('[1]')}: ${chalk.bold('https://github.com/angular/angular-cli/wiki/
 
     return {
       custom: program !== DEFAULT_PROGRAM,
-      protocol: 'http',
+      protocol: options.ssl ? 'https' : 'http',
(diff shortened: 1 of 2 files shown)


dwieeb:
I believe the Angular CLI now requires `--ssl-cert` and `--ssl-key` as of version 6, otherwise it appears to generate an SSL certificate at some unknown location and then the user is presented with this: 

image

I'd rather we document a proper strategy for HTTPS in our docs using a long-lived SSL certificate that is trusted by their system. I already started work on `ionic ssl generate`: https://beta.ionicframework.com/docs/cli/ssl-generate

dwieeb:
Subscribe to this issue for updates on HTTPS support: https://github.com/ionic-team/ionic-cli/issues/3305

dwieeb:
This is all possible today by manually passing the SSL args to the Angular CLI:

```
ionic serve -- --ssl --ssl-cert=/path/to/cert --ssl-key=/path/to/key
```

The issue I linked above is how Ionic can make it even easier.

rdlabo:
thanks! this is great!!

PR #3444, opened 2018-07-31
fix(cli-utils/project): pass the host address as it is

rdlabo:
Fixes https://github.com/ionic-team/ionic-cli/issues/3410

Change `await isHostConnectable('localhost', port);` to `await isHostConnectable(options.address, port);` .  I think this is consistent `findOpenIonicPorts()`'s address.

---
- cause of my error
'localhost' can't resolve ip-address.
reference: https://github.com/nodejs/node/issues/2237#issuecomment-124718020

- cause of https://github.com/ionic-team/ionic-cli/issues/3410#issuecomment-408444428
change address `192.168.254.58`, but connect check by `localhost`.


dwieeb:
This looks great. I also have some additional changes related to how the Ionic CLI chooses a port.

rdlabo:
`checking for open port` is great!! thanks!!

PR #3457, opened 2018-07-31
Added forgotten proxyConfig property

glebmachine:
You forgot about proxyConfig, which totally indispensable for development.


dwieeb:
Thanks @glebmachine!

Unfortunately there are quite a few options we're missing. I wasn't super thrilled with how people would need to copy config from their `dev-server` task to the `ionic-cordova-serve` task... but I think this is just how the Angular CLI wants it.

glebmachine:
Yeah, at least it allows to redefine configuration, especially for ionic
serve task
-- 
Gleb Mikheev
CTO

*Beta Digital Production*
Bersenevsky Lane 5 Building 1

Mobile:  +7 (916) 118-28-21
Telegram: https://t.me/glebmikheev

www.betaagency.ru
www.facebook.com/betadigitalproduction

Issue #3475, opened 2018-08-02
`ionic g` will not make folder?


rdlabo:
https://beta.ionicframework.com/docs/building/migration is written like structure.

```
src/
├── app/
　   ├── components/
　   ├── services/
　   └── pages/
```

But starter template and `ionic g` don't support this structure. I have not chased for a long time, what has become of this problem?

If Ionic team is waiting for implementation, I would like to challenge this. or If it keeps going, I would like to make option `ionic g -d`(--directory).

I believe that Ionic v3 structure supported many beginner's learning, this structure is great. so 
I also want to make it standard specification in v4.

thanks.

dwieeb:
Haha, I literally just made some updates to those docs: https://github.com/ionic-team/ionic-docs/commit/a9cf5f3f564c37d13d70a821858366fbf112d269

Basically, we don't want to **force** apps into a structure, we want Angular to lead the way in recommended project structure. See original issue for more info: https://github.com/ionic-team/ionic-cli/pull/3392

rdlabo:
Haha, Oh...It's good timing.

Sure. I understand. But distinguish between pages or components is too difficult for beginner. 
So I challenge -d option.(not force) Please review it if I make a draft.
thanks.

dwieeb:
I agree it is difficult for a beginner. Hopefully the file names will be enough. We add the `.page.ts` suffix:

```
❯ l src/app/home
total 40
drwxr-xr-x   7 dan  staff   224B Jul 25 17:09 .
drwxr-xr-x  22 dan  staff   704B Jul 30 16:49 ..
-rw-r--r--   1 dan  staff   477B Jul 25 15:54 home.module.ts
-rw-r--r--   1 dan  staff   275B Jul 25 15:54 home.page.html
-rw-r--r--   1 dan  staff     1B Jul 25 15:54 home.page.scss
-rw-r--r--   1 dan  staff   669B Jul 25 15:54 home.page.spec.ts
-rw-r--r--   1 dan  staff   179B Jul 25 15:54 home.page.ts
```

vs component:

```
❯ l src/app/blah
total 24
drwxrwxr-x   6 dan  staff   192B Jul 30 16:35 .
drwxr-xr-x  22 dan  staff   704B Jul 30 16:49 ..
-rw-rw-r--   1 dan  staff    24B Jul 30 16:35 blah.component.html
-rw-rw-r--   1 dan  staff     0B Jul 30 16:35 blah.component.scss
[...]

PR #3477, opened 2018-08-03
WIP: feat(command): add option -s for change project of type=angular structure.

rdlabo:
**This request not work tabs template**

@dwieeb How about this kind of implementation? This work `ionic start --type=angular -s=true` and `ionic g -s=true`. (I will use `ionic g -s`).
This implementation don't need to change starter package. and user can choice use `-s` or not. 
 
please tell me your opinion.

Try https://github.com/ionic-team/ionic-cli/issues/3475

Task: 
- can use tabs template.
- can use `ionic g -s`.


rdlabo:
move https://github.com/ionic-team/ionic-cli/pull/3542.

Issue #3069, opened 2018-04-07
Testing Ionic 4: Ionic lab display problem


peterennis:
**Description:**
Display does not look good.

**Steps to Reproduce:**
Create new project and run with lab

**Output:**
```
C:\ae\adaept.com\aeion4>ionic serve --lab --browser chrome
> ng serve --host=0.0.0.0 --port=8100 --progress=false
> ionic-lab http://localhost:8100 --host localhost --port 8200 --app-name aeion4 --app-version 0.2.1

[OK] Development server running!

     Lab: http://localhost:8200
     Local: http://localhost:8100
     External: http://172.21.32.243:8100
     DevApp: aeion4@8100 on Merry-Christmas

[INFO] Browser window opened to http://localhost:8200!

[ng] webpack: wait until bundle finished: /?ionicplatform=android
[ng] Date: 2018-04-07T20:33:27.870Z
[ng] Hash: f5f13334cda7e469e1ff
[ng] Time: 19214ms
[ng] chunk {inline} inline.bundle.js (inline) 3.85 kB [entry] [rendered]
[ng] chunk {main} main.bundle.js (main) 30.4 kB [initial] [rendered]
[ng] chunk {polyfills} polyfills.bundle.js (polyfills) 553 kB [initial] [rendered]
[ng] chunk {styles} styles.bundle.js (styles) 49.2 kB [initial] [rendered]
[ng] chunk {vendor} vendor.bundle.js (vendor) 10.6 MB [initial] [rendered]
[ng] webpack: Compiled successfully.

C:\ae\adaept.com\aeion4>run

C:\ae\adaept.com\aeion4>ionic serve --lab --browser chrome
> ng serve --host=0.0.0.0 --port=8101 --progress=false
> ionic-lab http://localhost:8101 --host localhost --port 8201 --app-name aeion4 --app-version 0.2.1

[OK] Development server running!

     Lab: http://localhost:8201
     Local: http://localhost:8101
     External: http://172.21.32.243:8101
     DevApp: aeion4@8101 on Merry-Christmas

[INFO] Browser window opened to http://localhost:8201!

[ng] webpack: wait until bundle finished: /?ionicplatform=android
[ng] webpack: wait until bundle finished: /?ionicplatform=ios&ionicstatusbarpadding=true
[ng] Date: 2018-04-07T20:34:39.403Z
[ng] Hash: f5f13334cda7e469e1ff
[ng] Time: 19315ms
[...]

PR #3766, opened 2018-11-28
fix(lab): remove css that places the statusbar above the header

brandyscarney:
fixes android and windows statusbar so it is on top of the header in lab

references #3069 

before:
localhost_8200_

aftah:
localhost_8200_ 1


brandyscarney:
Note that this _technically_ will break Android on Chrome in v3, but it's already broken in iOS in Chrome since before this PR, and the fix will need to be in v3. I don't think this PR breaks it so much as it exposes that forced statusbar padding is now broken in latest Chrome.

PR #3809, opened 2019-01-04
fix(): ionic doctor ViewportFixNotSet Ailment

reajuria:
Regex doesn't match on a multiline meta tag, an strict matcher is proposed to prevent match on other meta tags, only should match if a meta tag has name="viewport" property

done in test bench: https://regex101.com/r/yy1zT2/1


dwieeb:
@reajuria I don't think that's working. It's not finding the other valid viewport settings in your test bench. 

I'm more comfortable with just accepting characters that aren't a closing bracket (`>`), such as this regex:

```
\<meta[^>]*?name="viewport"[^>]*?viewport-fit=cover
```

dwieeb:
Ah, nevermind, I was doing that wrong. Sorry, never used that website.

dwieeb:
Thanks @reajuria!! Great work.

Issue #3428, opened 2018-07-27
`ionic build --cordova` with option to exclude bundled Cordova assets


dwieeb:
Right now `ionic build --cordova` will build web assets for Cordova, bundling `cordova.js`, all the plugin JS, and adding `cordova.js` to `index.html`. 

Pro's Deploy service never needs to send bundled Cordova JS files during a deploy because they won't change unless native code changes, in which case Deploy cannot be used.

There should be an option to exclude Cordova assets during a web asset build for Cordova.

possible UX examples:

```
ionic build --deploy
```

```
ionic build --cordova --no-cordova-assets
```

PR #3830, opened 2019-01-28
feat(serve): add --devapp support for Ionic 4 projects

tlancina:
Resolves https://github.com/ionic-team/ionic-cli/issues/3428


dwieeb:
While DevApp may have picked up Ionic 4 apps prior to this PR being merged, these changes will enable Cordova functionality via the new `--cordova-mock` option of the Angular Toolkit: https://github.com/ionic-team/angular-toolkit/commit/a6596367653802ab44fb499f2583d3caf4e7137e

Once the mock `cordova.js` is in an app's `index.html`, DevApp will shim the request with a proper `cordova.js` file and load the DevApp plugins into Ionic app.

dwieeb:
This PR will close https://github.com/ionic-team/ionic-cli/issues/3428, because the `--cordova-mock` option is also supported during `ionic build`:

```
ionic build --cordova --cordova-mock
```

This will perform a web asset build for Cordova with the following adjustments:
- `cordova.js`, `cordova_plugins.js`, and all Cordova plugin JS are excluded
- A mock `cordova.js` file is included in the build, existing as a script in `index.html` and in the resultant build files (`www/cordova.js`)

Because the mock script is included in `www/`, one may need to `rm www/cordova.js` after the build if so desired.

Additionally, because of how the Angular CLI appends hashes during prod builds, the actual name of the script may change, e.g. `cordova.abcd1234.js`.

tlancina:
@dwieeb thinking about it, does #3428 work with ```ionic build --cordova --no-cordova-assets``` as well? 

I just updated the tests to reflect that the default value for `devapp` from `ServeRunner.createOptionsFromCommandLine` will be false if undefined, but if this isn't intended then we should talk about it.  The default value from the command metadata means when run as an actual command it will get the correct default value, but in the tests there are no defaults coming from metadata ([ionic1/\_\_tests\_\_/serve.ts#L35](https://github.com/ionic-team/ionic-cli/blob/dbd912f5fec0b325cfceadd971ca986706ef8db6/packages/ionic/src/lib/project/ionic1/__tests__/serve.ts#L35) vs. [ionic1/serve.ts#L66](https://github.com/ionic-team/ionic-cli/blob/dbd912f5fec0b325cfceadd971ca986706ef8db6/packages/ionic/src/lib/project/ionic1/serve.ts#L66)).

packages/ionic/src/lib/project/ionic-angular/serve.ts
```diff
@@ -97,6 +105,7 @@ export class IonicAngularServeRunner extends ServeRunner<IonicAngularServeOption
       livereloadPort,
       notificationPort,
       env: options['env'] ? String(options['env']) : undefined,
+      devapp: !!options.devapp,
```
dwieeb: I don't think this is necessary here

Issue #3856, opened 2019-02-08
v1-toolkit support for CORS


waded:
While `angular` project resources served using `ionic serve` are sent with header `Access-Control-Allow-Origin: *` (as `ng serve` would), `ionic1` project resources served via v1-toolkit are not.

We'd like to see `Access-Control-Allow-Origin: *` sent by v1-toolkit as well.

This header matters when `ionic serve` is used to dev/test Ionic projects where the index.html is filtered/generated with a different origin than the remaining resources. E.g. index.html is generated and served from http://localhost:8080 but the remainder of the resources are served using `ionic serve` at http://localhost:8100. In this case allowing the cross-origin request allows the Ionic application to successfully XHR .tpl.html files. Without CORS allowing GET, those text/html responses are subject to CORB, resulting in empty body on those requests.

The scenario that resulted in `ng serve` supporting this is https://github.com/angular/angular-cli/issues/2872.

Given the dev/test purpose of Ionic CLI, and given `ng serve` supports it, it seems reasonable to add the same header for Ionic 1. We're moving to Ionic 4, but doing so means working in a mixed-version environment for some time. 

I will submit a PR that shows exactly what I mean. It'll be similar to https://github.com/angular/angular-cli/commit/7c834a87415e2d5201834c4afdd7b43c71d5f0f0.

PR #3857, opened 2019-02-08
v1-toolkit now sends Access-Control-Allow-Origin: *

waded:
This is the change I'd make for #3856. We've tested this change and it works for the scenario described, as compared to ionic@4.10.2 where .tpl.html requests fail.

### DIFF
--- a/packages/@ionic/v1-toolkit/src/lib/serve.ts
+++ b/packages/@ionic/v1-toolkit/src/lib/serve.ts
@@ -128,6 +128,10 @@ async function createHttpServer(options: ServeOptions): Promise<ζexpress.Applic
   };
 
   app.get('/', serveIndex);
+  app.use('/', (req, res, next) => {
+    res.header('Access-Control-Allow-Origin', '*');
+    next();
+  });
   app.use('/', express.static(options.wwwDir));
 
   // Cordova


Matttaylor8910:
This looks great! I need this in one of my projects as well for a specific dev scenario.

dwieeb:
Do you think a better solution would be to use the CORS middleware? https://expressjs.com/en/resources/middleware/cors.html

waded:
@dwieeb I considered that and even implemented & tested with that, but it was overkill in our scenario. We did not e.g. need OPTIONS support.

waded:
@dwieeb (I also noted that angular-cli did not do anything more than add the single header, and I did not see anyone asking for more than what they'd done for `ng serve`.)

waded:
That said, if you're OK w/ v1-toolkit taking on those dependencies, I'd be happy to resubmit this w/ usage of the cors middleware instead.

dwieeb:
@waded All good, just wanted your thoughts on it.

I'm not sure why CI is failing. Might be a typescript 3.3.3 issue.

dwieeb:
Thanks @waded! This is great work 🍻

waded:
Thanks for getting this in! I'll leave #3856 be for you to close once this releases or when you see fit.

PR #3895, opened 2019-03-06
feat(integrations): add ability to pass web-dir option to capacitor

jthoms1:
This PR includes code changes to the integration command for capacitor.  It adds the ability to pass a new parameter called web-dir.  This is then passed on capacitor if it exists.


packages/ionic/src/lib/integrations/capacitor/index.ts
```diff
@@ -24,13 +26,22 @@ export class Integration extends BaseIntegration {
       if (details.enableArgs[1]) {
         packageId = details.enableArgs[1];
       }
+
+      /*
+       * Use parseArgs on the process variables because integrations enable has its own set of options outside
+       * of the options that we will pass to capacitor
+       */
+      const parsedArgs = parseArgs(process.argv.slice(4));
```
dwieeb: Why not parse `details.enableArgs`?

jthoms1: Because the options are not available since it is not defined as an actual option to the command.

jthoms1: Ok, updated based on recommendation

Issue #3169, opened 2018-05-16
Print console.log and console.error messages in terminal for Angular


dwieeb:
This has to be re-implemented for use with the Angular CLI. We'd need to inject a client JS file into the Angular dev server and connect to a websocket server running from the Ionic CLI.

dwieeb:
@GuilhermeBCC Can you explain why these logs are vital for your tests? What sort of testing are you doing?

GuilhermeBCC:
Well, when I'm running on the device I do not even see when there are any errors, the application simply hangs, there is no information to aid the fix. I am even stopped now because there is an error that I can not trace.

dwieeb:
If the webview is loaded, you should be able to attach dev tools to debug issues and see console logs:

iOS: https://beta.ionicframework.com/docs/building/ios#using-safari-web-inspector
Android: https://beta.ionicframework.com/docs/building/android#using-chrome-devtools-to-debug

Borongaj:
Dev tools are nice, but we have to restart manually everytime we build the app. Sometimes the dev tools window closes immediatly after launch.

Is it normal that Ionic 3 projects don't show logs anymore too ?

obkdev:
I'm not able to view the console logs if the app is pushed to my device via Ionic Pro... How do I see logs  of a 'deployed' app?? 

Please help! Thank you!

obkdev:
How do I see logs for an app that is installed on the device via Ionic Pro? I'm not able to inspect a 'deployed' app in the same way I'm able to inspect an app I run on the device via 'ionic cordova run android' .. Please can you suggest a solution? 

Thank you!

sjregan:
It would great to have this feature back again. The Safari Debugger is very flakey, it often crashes, doesn't detect device unless you restart. It's much nicer to be able to have a dedicated window on screen that always displays the log output.

Borongaj:
Recently, since @ionic/angular 4.0.1 or 4.0.2, the Safari debugger does not show Ionic logs anymore...

ludufre:
Is there any forecast?

timbru31:
Given that it's nearly 1 year without any progress, I believe you can't expect a forecast/ETA/release date

alejandrosklar:
> @GuilhermeBCC Can you explain why these logs are vital for your tests? What sort of testing are you doing?

[...]

PR #3900, opened 2019-03-08
feat(serve): support --consolelogs for cordova-serve in angular projects

tlancina:
Anything that triggers an `ionic-cordova-serve`, as defined in
`@ionic/angular-toolkit`, will pass through the `--consolelogs` and
`--consolelogs-port` flags, allowing you to print app console logs to
the terminal. Right now that is `ionic cordova run ios --livereload`
and `ionic serve --devapp`,

Addresses https://github.com/ionic-team/ionic-cli/issues/3169.

Depends on https://github.com/ionic-team/angular-toolkit/pull/100.


packages/ionic/src/lib/project/angular/serve.ts
```diff
@@ -85,9 +102,16 @@ The dev server can use HTTPS via the ${chalk.green('--ssl')} option ${chalk.bold
     const ssl = options['ssl'] ? Boolean(options['ssl']) : undefined;
     const configuration = options['configuration'] ? String(options['configuration']) : (prod ? 'production' : undefined);
     const sourcemaps = typeof options['source-map'] === 'boolean' ? Boolean(options['source-map']) : undefined;
+    const consolelogs = typeof options['consolelogs'] === 'boolean' ? Boolean(options['consolelogs']) : undefined;
+    let consolelogsPort;
+    if (consolelogs) {
+      consolelogsPort = str2num(options['consolelogs-port'], DEFAULT_CONSOLE_LOGS_PORT);
```
dwieeb: Why not just:

```ts
const consolelogsPort = consolelogs ? str2num(options['consolelogs-port'], DEFAULT_CONSOLE_LOGS_PORT) : undefined;
```

tlancina: Ah good call, that block went through several incarnations as I was testing out the various args combos and how it was printed by the ng cli.

PR #3951, opened 2019-04-19
fix(react/vue): add project type at creat project.

rdlabo:
Change be able to valid project type react (and vue).

related: https://github.com/ionic-team/ionic-cli/pull/3936


dwieeb:
cc @jthoms1

PR #3963, opened 2019-05-01
fix(capacitor): Send --npm-client param to capacitor init

jcesarmobile:
On Capacitor beta 22, it added support for yarn package manager.
If it detects yarn installed and no `--npm-client` param is passed, it will prompt the user to choose a package manager, but that blocks the ionic cli. (see https://github.com/ionic-team/capacitor/issues/1453)

This PR will pass the `--npm-client` param to Capacitor CLI based on the configured `npmClient`

### DIFF
--- a/packages/ionic/src/lib/integrations/capacitor/index.ts
+++ b/packages/ionic/src/lib/integrations/capacitor/index.ts
@@ -36,6 +36,7 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
       if (parsedArgs['web-dir']) {
         options.push('--web-dir', parsedArgs['web-dir']);
       }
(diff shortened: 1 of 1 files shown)


dwieeb:
@jcesarmobile Can we also disable prompts in the Capacitor CLI if it's not a TTY? e.g. `process.stdin.isTTY` should be `false` when the Capacitor CLI is used as a subprocess.

jcesarmobile:
In that case, what should happen? Fail because the npm client wasn’t provided? Or set it to npm as default?
I think it already disables some prompts, but not sure how to properly test it

dwieeb:
Probably fail with exit code >0, yes.

You can test it like this:

```
npx cap blah blah | cat
```

PR #3995, opened 2019-05-30
chore(): support consolelogs in native run

mhartington:
Adds back console logs support to native run.


dwieeb:
This solution will forward all used ports by a serve runner: https://github.com/ionic-team/ionic-cli/commit/d2e0c81c7e15d3f10cb112a0f8ae8258ae39b6cb

dwieeb:
I see you used findClosestOpenPort for the consolelogs port... we may want that.

PR #4020, opened 2019-06-11
fix(react/vue): prevent usage of Cordova on vue and react projects

jcesarmobile:
Prevents users from adding Cordova to vue and react projects.

Prevents it for `ionic integrations enable cordova` and ` ionic cordova platform add`, not sure if more commands need it too.

### DIFF
--- a/packages/ionic/src/lib/integrations/cordova/index.ts
+++ b/packages/ionic/src/lib/integrations/cordova/index.ts
@@ -23,7 +23,10 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
 
   async add(details: IntegrationAddDetails): Promise<void> {
     if (this.e.project.type === 'vue' || this.e.project.type === 'react') {
-      throw new FatalException(`Cordova is not supported for ${this.e.project.type} projects`);
+      throw new FatalException(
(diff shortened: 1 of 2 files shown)


packages/ionic/src/lib/integrations/cordova/index.ts
```diff
@@ -21,6 +22,9 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
   }
 
   async add(details: IntegrationAddDetails): Promise<void> {
+    if (this.e.project.type === 'vue' || this.e.project.type === 'angular') {
```
dwieeb: Did you mean `react` here?

brandyscarney:
It looks good! I think we should also link off to some documentation with the warning though.

packages/ionic/src/lib/integrations/cordova/index.ts
jcesarmobile: wops, fixed

jcesarmobile:
@brandyscarney linked the overview pages that will contain the information once https://github.com/ionic-team/ionic-docs/pull/744 gets merged.

PR #4039, opened 2019-06-18
ionic deploy add/configure

giocalitri:


packages/ionic/src/commands/deploy/configure.ts
```diff
@@ -0,0 +1,70 @@
...
+      summary: 'Overrides the Deploy plugin (cordova-plugin-ionic) configuration in the project',
+      description: `
+This command overrides the Deploy plugin (cordova-plugin-ionic) configuration Capacitor projects.
+
+In a Capacitor project, if the plugin is already installed, it overrides the configuration variables in the native projects.
+
+For a Cordova project this is not implemented because it is better to reinstall the plugin with the different
+paramenters and let Cordova deal with the changes.
+      `,
+      footnotes: [
```
dwieeb: You're not using these footnotes in the description.

packages/ionic/src/commands/deploy/add.ts
```diff
@@ -0,0 +1,104 @@
...
+      summary: 'Adds the Deploy plugin (cordova-plugin-ionic) to the project',
+      description: `
+This command adds the Deploy plugin (cordova-plugin-ionic) for both Cordova and Capacitor projects.
+
+For a Condova project it just takes care of running the proper cordova cli command with the submitted parameters.
+
+For a Capacitor project it runs all the steps necessary to install the plugin, sync with the native projects and
+add the configuration to the proper iOS and Android configuration files.
+      `,
+      footnotes: [
```
dwieeb: You're not using these footnotes in the description.

packages/ionic/src/commands/deploy/core.ts
```diff
@@ -0,0 +1,374 @@
...
+
+  protected async preRunCheckInputs(options: CommandLineOptions): Promise<void> {
+    const updateMethodList: string[] = ['auto', 'background', 'none'];
+    const defaultUpdateMethod = 'background';
+    // handle the app-id option in case the user wants to override it
+    if (!options['app-id']) {
+      const appId = await this.getAppId();
+      if (!appId) {
+        this.env.log.warn(
+          `No app id found on the project: consider running ${input('ionic link')} to Connect local apps to Ionic`
```
dwieeb: Capitalize `id`

packages/ionic/src/commands/deploy/core.ts
```diff
@@ -0,0 +1,374 @@
...
+        validate: v => validators.required(v),
+      });
+    }
+
+    // validate that the update-method is allowed
+    let overrideUpdateMethodChoice = false;
+    if (options['update-method'] && !updateMethodList.includes(options['update-method'] as string)) {
+      if (this.env.flags.interactive) {
+        this.env.log.nl();
+        this.env.log.warn(`${strong(options['update-method'] as string)} is not a valid update method; choose a valid one`);
```
dwieeb: Please use the `input` color. update-method is a user input. 

Additionally, in other places in help documentation and messages, we prefix these options in prose with `--` so users know what we're referring to, even for prompted options.

```ts
strong(options['update-method'] as string) -> input(`--${options['update-method']}`)
```

packages/ionic/src/commands/deploy/core.ts
```diff
@@ -0,0 +1,374 @@
...
+        default: defaultUpdateMethod,
+        validate: v => validators.required(v) && contains(updateMethodList, {})(v),
+      });
+    }
+
+    // check advanced options if present
+    if (options['max-store'] && validators.numeric(options['max-store'] as string) !== true) {
+      if (this.env.flags.interactive) {
+        this.env.log.nl();
+        this.env.log.warn(`${strong(options['max-store'] as string)} is not a valid Max Store value; please specify an integer`);
```
dwieeb:
```ts
strong(options['max-store'] as string) -> input(`--${options['max-store']}`)
```

packages/ionic/src/commands/deploy/core.ts
```diff
@@ -0,0 +1,374 @@
...
+        name: 'max-store',
+        message: `Max Store:`,
+        validate: v => validators.required(v) && validators.numeric(v),
+      });
+    }
+
+    if (options['min-background-duration'] && validators.numeric(options['min-background-duration'] as string) !== true) {
+      if (this.env.flags.interactive) {
+        this.env.log.nl();
+        this.env.log.warn(`${strong(options['min-background-duration'] as string)} is not a valid Min Background Duration value; please specify an integer`);
```
dwieeb:
```ts
strong(options['min-background-duration'] as string) -> input(`--${options['min-background-duration']}`)
```

packages/ionic/src/commands/deploy/core.ts
```diff
@@ -0,0 +1,374 @@
...
+        type: 'input',
+        name: 'min-background-duration',
+        message: `Min Background Duration:`,
+        validate: v => validators.required(v) && validators.numeric(v),
+      });
+    }
+    if (options['update-api'] && validators.url(options['update-api'] as string) !== true) {
+      if (this.env.flags.interactive) {
+        this.env.log.nl();
+        this.env.log.warn(`${strong(options['update-api'] as string)} is not a valid Update Api value; please specify a url`);
```
dwieeb:
```ts
strong(options['update-api'] as string) -> input(`--${options['update-api']}`)
```

packages/ionic/src/commands/deploy/index.ts
```diff
@@ -9,7 +9,7 @@ export class DeployNamespace extends Namespace {
       name: 'deploy',
       summary: 'Appflow Deploy functionality',
       description: `
-Interface to execute commands about deploy builds on Ionic Appflow.
+Interface to execute commands about deploy plugin in a project and deploy builds on Ionic Appflow.
```
dwieeb: This reads poorly IMO. Can we say something like:

```
These commands integrate with Ionic Appflow to configure the deploy plugin in your project and run remote builds.
```

packages/ionic/src/commands/deploy/add.ts
```diff
@@ -0,0 +1,104 @@
...
+    const dashUrl = this.env.config.getDashUrl();
+
+    return {
+      name: 'add',
+      type: 'project',
+      summary: 'Adds the Deploy plugin (cordova-plugin-ionic) to the project',
+      description: `
+This command adds the Deploy plugin (cordova-plugin-ionic) for both Cordova and Capacitor projects.
+
+For a Condova project it just takes care of running the proper cordova cli command with the submitted parameters.
```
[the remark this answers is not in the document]
nphyatt: Also Cordova not Condova

packages/ionic/src/commands/deploy/add.ts
```diff
@@ -0,0 +1,104 @@
...
+        outputOptions.push(`--variable`);
+        outputOptions.push(`${cordovaKey}=${options[optionKey]}`);
+      }
+    }
+    return outputOptions;
+  }
+
+  async preRun(inputs: CommandLineInputs, options: CommandLineOptions): Promise<void> {
+    // check if it is already installed
+    const alreadyAdded = await this.checkDeployInstalled();
```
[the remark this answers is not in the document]
nphyatt: cc @dwieeb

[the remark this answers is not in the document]
dwieeb: It's a centralized way to statefully represent what's integrated with your app. If you can add and remove something from your app, and it's not fully represented by some other system, e.g. `package.json` and installing Cordova plugins/platforms, then it's probably an integration.

If we think that the Deploy integration is fully represented by simply having the deploy plugin installed, then that's fine. If there's any other piece we need, for example bootstrapping something in `app.module.ts`, we may want to consider making it an integration.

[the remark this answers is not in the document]
nphyatt: I think for now I'm of the opinion that it's represented by the package.json particularly because you can remove it using npm or cordova cli too with no opportunity for for the CLI to detect it and currently that's actually what the docs represent. That might change overtime I suppose but traditionally it has been a cordova plugin managed by cordova similar to how you manage other plugins and now a capacitor plugin. I think the command it really a utility only command...a helper for making the plugin installation and configuration easier by helping automate the official install/management tools (cordova CLI/ npm/config.xml/Info.plist).

packages/ionic/src/commands/deploy/core.ts
```diff
@@ -316,7 +325,7 @@ export abstract class DeployConfCommand extends Command {
     if (options['update-method'] && !updateMethodList.includes(options['update-method'] as string)) {
       if (this.env.flags.interactive) {
         this.env.log.nl();
-        this.env.log.warn(`${strong(options['update-method'] as string)} is not a valid update method; choose a valid one`);
+        this.env.log.warn(`--${input(options['update-method'] as string)} is not a valid update method; choose a valid one`);
```
dwieeb: This makes the colors mismatch between the two dashes and the option name.

nphyatt:
@dwieeb We everything looking ok now?

PR #4050, opened 2019-06-25
Added platform to deploy configure

giocalitri:


packages/ionic/src/commands/deploy/configure.ts
```diff
@@ -41,6 +49,10 @@ For Cordova projects this is not implemented because it is better to reinstall t
     }
     // check if there are native integration installed
     await this.requireNativeIntegration();
+    // check that if an input is provided, it is valid
+    if (inputs[0] && !['ios', 'android'].includes(inputs[0])) {
+      throw new FatalException('Only `ios` or `android` can be used.');
```
dwieeb: Please wrap `android` and `ios` in `input` color (no need for backticks).

packages/ionic/src/commands/deploy/configure.ts
```diff
@@ -22,11 +22,19 @@ For Cordova projects this is not implemented because it is better to reinstall t
...
-        '--app-id="abcd1234" --channel-name="Master" --update-method="background" --max-store=2 --min-background-duration=30',
+        '--app-id=abcd1234 --channel-name=Master --update-method=background --max-store=2 --min-background-duration=30',
+        'android',
+        'ios',
       ],
       options: this.commandOptions,
+      inputs: [
+        {
+          name: 'platform',
+          summary: `The native platform (e.g. ${input('ios | android')})`,
```
dwieeb: `${['ios', 'android'].map(v => input(v)).join(', ')}` is what we do in other places when listing common values

PR #4065, opened 2019-07-03
Added support for custom Capacitor build directory for ionic deploy manifest

giocalitri:


packages/ionic/src/commands/deploy/manifest.ts
```diff
@@ -31,10 +32,32 @@ export class DeployManifestCommand extends Command {
       throw new FatalException(`Cannot run ${input('ionic deploy manifest')} outside a project directory.`);
     }
 
-    const buildDir = path.resolve(this.project.directory, 'www'); // TODO: this is hard-coded
+    let buildDir: string;
+    if (this.project.getIntegration('capacitor') !== undefined) {
```
dwieeb: Don't you already do this logic in `deploy/core.ts`? Can you reuse it? If you need to, you can create another base class for the deploy commands that they all inherit from.

packages/ionic/src/commands/deploy/manifest.ts
```diff
@@ -31,10 +32,32 @@ export class DeployManifestCommand extends Command {
...
+      const conf = new CapacitorConfig(path.resolve(this.project.directory, CAPACITOR_CONFIG_FILE));
+      const webDir = conf.get('webDir');
+      if (webDir) {
+        buildDir = webDir;
+      } else {
+        throw new FatalException(`WebDir parameter has no value set in the Capacitor config file ${input(CAPACITOR_CONFIG_FILE)}`);
+      }
+    } else if (this.project.getIntegration('cordova') !== undefined) {
+      // for cordova it is hardcoded because www is mandatory
+      buildDir = path.resolve(this.project.directory, 'www');
```
dwieeb: This should use `await this.project.getDistDir()`

packages/ionic/src/commands/deploy/manifest.ts
```diff
@@ -31,10 +32,32 @@ export class DeployManifestCommand extends Command {
...
 
-    const buildDir = path.resolve(this.project.directory, 'www'); // TODO: this is hard-coded
+    let buildDir: string;
+    if (this.project.getIntegration('capacitor') !== undefined) {
+      const conf = new CapacitorConfig(path.resolve(this.project.directory, CAPACITOR_CONFIG_FILE));
+      const webDir = conf.get('webDir');
+      if (webDir) {
+        buildDir = webDir;
+      } else {
+        throw new FatalException(`WebDir parameter has no value set in the Capacitor config file ${input(CAPACITOR_CONFIG_FILE)}`);
```
dwieeb: Use _exact_ names for properties. I didn't know what a "WebDir parameter" was. Try this:

```
The ${input('webDir')} property must be set in the Capacitor configuration file (${input(CAPACITOR_CONFIG_FILE)}).

See the Capacitor docs for more information: ${strong('https://capacitor.ionicframework.com/docs/basics/configuring-your-app')}
```

packages/ionic/src/commands/deploy/manifest.ts
nphyatt: @dwieeb does `this.project.getDistDir()` return the correct dir for capacitor apps as well? If so I think this entire PR can be boiled down to that one line.

nphyatt: it would appear not

nphyatt: so maybe we can move all of this code into the `getDistDir` function itself?

dwieeb: Yeah, probably. 

I would personally like to see more of what has been accomplished in `ionic deploy` put into lower-level code so it can be reused.

More examples:

- Projects will only be integrated with Cordova _or_ Capacitor (or neither), so that warrants a method in the `Project` class for getting the native integration (currently this exists in `deploy/core.ts`). 
- The Capacitor native project configuration logic (Info.plist and strings.xml manipulation) should probably be abstracted and put into `src/lib`. It's best to keep code in `src/commands` as lean as possible.

nphyatt: Cool maybe we can make a separate refactor issue for future refactoring.

packages/ionic/src/commands/deploy/manifest.ts
```diff
@@ -31,10 +32,24 @@ export class DeployManifestCommand extends Command {
...
 
-    const buildDir = path.resolve(this.project.directory, 'www'); // TODO: this is hard-coded
+    const integration = await this.getAppIntegration();
+    let buildDir: string;
+    if (integration && ['capacitor', 'cordova'].includes(integration)) {
+      buildDir = await this.project.getDistDir();
+    } else {
+      throw new FatalException(
+        `It looks like your app isn't integrated with Capacitor or Cordova.\n` +
+        `In order to generate a manifestfor the Appflow Deploy plugin, you will need to integrate your app with Capacitor or Cordova. See the docs for setting up native projects:\n\n` +
```
dwieeb: typo: manifestfor -> manifest for

packages/ionic/src/lib/project/index.ts
```diff
@@ -520,7 +521,20 @@ export abstract class Project implements IProject {
...
 
   async getDistDir(): Promise<string> {
-    return path.resolve(this.directory, 'www');
+    if (this.getIntegration('capacitor') !== undefined) {
+      const conf = new CapacitorConfig(path.resolve(this.directory, CAPACITOR_CONFIG_FILE));
+      const webDir = conf.get('webDir');
+      if (webDir) {
+        return path.resolve(this.directory, webDir);
+      } else {
+        throw new FatalException(
```
dwieeb: Would it be better to warn and default to `www` here, instead?

packages/ionic/src/commands/deploy/manifest.ts
```diff
@@ -31,10 +32,24 @@ export class DeployManifestCommand extends Command {
...
+        `Android: ${strong('https://ionicframework.com/docs/building/android')}\n`
+      );
+    }
+
     const manifest = await this.getFilesAndSizesAndHashesForGlobPattern(buildDir);
 
-    await writeFile(path.resolve(buildDir, 'pro-manifest.json'), JSON.stringify(manifest, undefined, 2), { encoding: 'utf8' });
+    const manifestPath = path.resolve(buildDir, 'pro-manifest.json');
+    await writeFile(manifestPath, JSON.stringify(manifest, undefined, 2), { encoding: 'utf8' });
+    this.env.log.ok(`Manifest ${input(manifestPath)} correctly generated\n`);
```
dwieeb: This will print an absolute path, which aren't used often in CLI output. Instead, use `prettyPath`.

```ts
this.env.log.ok(`Appflow Deploy manifest written to ${input(prettyPath(manifestPath))}!`);
```

packages/ionic/src/commands/deploy/manifest.ts
```diff
@@ -31,10 +32,24 @@ export class DeployManifestCommand extends Command {
       throw new FatalException(`Cannot run ${input('ionic deploy manifest')} outside a project directory.`);
     }
 
-    const buildDir = path.resolve(this.project.directory, 'www'); // TODO: this is hard-coded
+    const integration = await this.getAppIntegration();
+    let buildDir: string;
+    if (integration && ['capacitor', 'cordova'].includes(integration)) {
+      buildDir = await this.project.getDistDir();
+    } else {
+      throw new FatalException(
```
dwieeb: This message is still written in two places. Why can't you reuse it from `deploy/core.ts`?

packages/ionic/src/lib/project/index.ts
giocalitri: well, this I do not know. It depends on how we want the code to behave: if we prefer that people fix the project setting the proper variables or we prefer to make assumptions.
I do not have strong feelings here, but I believe it is more a general product decision.

jcesarmobile: I don't think Capacitor commands will work if there is no webDir configured, it already works by setting it to www as default. So this could only happen if users manually delete it, and will break everything, so I like the FatalException here.

PR #4083, opened 2019-07-14
Update updates.ts

Tommertom:
Many people get confused with CLI update vs Framework update. See forum. Maybe this avoids frustration and put-offs.


dwieeb:
@Tommertom Thanks. The confusion of Ionic CLI vs Framework has been a problem forever. Aside from changing the package name to `@ionic/cli`, I'm not sure what else we can do here.

PR #4086, opened 2019-07-15
docs(cordova resources): bad angles encoding

rtpHarry:
Please, can somebody confirm that I have done the correct notation?

I've turned `>` into `&gt;` 

I looked around but couldn't find any exact matches for the syntax.

### DIFF
--- a/packages/@ionic/cli-framework/package.json
+++ b/packages/@ionic/cli-framework/package.json
@@ -62,7 +62,7 @@
     "@types/write-file-atomic": "^2.1.1",
     "jest": "^23.0.1",
     "jest-cli": "^24.1.0",
-    "lint-staged": "^8.0.0",
+    "lint-staged": "^9.2.0",
     "ts-jest": "^23.10.1",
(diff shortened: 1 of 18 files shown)


dwieeb:
I assume this PR is for fixing the docs on the website: https://ionicframework.com/docs/cli/commands/cordova-resources

The text you're updating is also used in the output of `ionic help`, and it'll show up there as `&lt;platform&gt;`.

This will have to be fixed in our cli-scripts: https://github.com/ionic-team/ionic-cli/blob/develop/packages/cli-scripts/src/docs/index.ts (run `npm run docs`).

Let me know if you want to continue or not. I appreciate it either way. 😄

rtpHarry:
@dwieeb yeah I've let nearly a week go by and I haven't had a chance to look into what you've said. Please feel free to take this over. One parting note though - the snippet I'm "fixing" isn't actually trying to show a html tag, it was just somebody using it as `<replacethis>` notation, so maybe there is are better wrapper characters to use?

PR #4119, opened 2019-08-15
fix(react): Disable CRA from opening window so two windows don't open

elylucas:


packages/ionic/src/lib/project/react/serve.ts
```diff
@@ -137,21 +137,25 @@ export class ReactServeCLI extends ServeCLI<ReactServeOptions> {
...
-    if (options.open !== true) {
-      envVars.BROWSER = 'none';
+    envVars.BROWSER = 'none';
+
+    /*
+      By default, CRA binds to localhost,
+      but if you specify it, it puts a warning in the console,
+      so don't set the HOST if the address is set to 'localhost'
+    */
+    if (options.address !== 'localhost') {
```
dwieeb: Please use `DEFAULT_ADDRESS`

Issue #4118, opened 2019-08-15
`ionic start` should ask for the framework


lcswillems:
It would be nice if `ionic start` asks for the framework we want to use (angular, react or vue)

dwieeb:
@lcswillems We had planned on this when Ionic React reaches 1.0 final, but I see no reason why we can't do something like:

Choose your framework:
- Angular (stable)
- React (RC)
- Vue (experimental)

lcswillems:
I think it would be a good idea! It would make creation of projects easier, because every time, I have to search the doc to know the good parameter name and value.

PR #4161, opened 2019-09-20
feat(): add prompt for framework

mhartington:
Add prompt for picking different frameworks.

```
ionic start myApp
ionic start myApp tabs
ionic start myApp tabs --type <angular,react>
```

resolves https://github.com/ionic-team/ionic-cli/issues/4118


packages/ionic/src/lib/start.ts
```diff
@@ -194,7 +194,16 @@ export async function getStarterList(config: IConfig, tag = 'latest'): Promise<S
 export function getStarterProjectTypes(): string[] {
   return lodash.uniq(STARTER_TEMPLATES.map(t => t.projectType));
 }
-
+export const SUPPORTED_FRAMEWORKS = [
+  {
+    name: 'angular',
```
dotNetkow: minor: Capitalize both framework names?

packages/ionic/src/commands/start.ts
```diff
@@ -304,6 +304,36 @@ Use the ${input('--type')} option to start projects using older versions of Ioni
     }
   }
 
+  async getProjectType() {
+    if (this.env.flags.interactive) {
```
dwieeb: With a default provided, you shouldn't need to do this. You can confirm by removing the `else` and running with `--no-interactive`

dwieeb:
@mhartington looks like some lint issues in CI

Issue #4156, opened 2019-09-13
ionic cordova build iOS hangs only when iphone attached to Mac


rolinger:
**Description:**
Using cli 5.2.6, running Ionic v1 app - running `ionic cordova build iOS` hangs the app after this file is compiled ONLY if an iPhone is attached to the Mac.  Disconnect the iPhone and it compiles just fine.  Running `cordova build iOS` builds just fine even with an iPhone connected to the Mac.

`platforms/ios/build/sharedpch/SharedPrecompiledHeaders/6268428044778614059/MyApp-Prefix.pch.dia`

<edit correction>

I suspect it "might" be related to `Cordova-plugin-firebasex` - this is happening with firebasex plugin versions 6x and higher.  I brought this to the attention of the plugin owner/developer (Dave Allen) and he believes its a ionic cordova build wrapper issue as the issue doesn't happen when just doing `cordova build iOS`.

My Environment:

```
Ionic:

   Ionic CLI         : 5.2.6 (/usr/local/lib/node_modules/ionic)
   Ionic Framework   : ionic1 1.0.0
   @ionic/v1-toolkit : 1.0.22

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : ios 5.0.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.1.3, (and 32 other plugins)

Utility:

   cordova-res : 0.6.0 
   native-run  : 0.2.8 

System:

   ios-deploy : 1.9.4
   ios-sim    : 8.0.2
   NodeJS     : v10.16.3 (/usr/local/bin/node)
   npm        : 6.11.2
   OS         : macOS High Sierra
   Xcode      : Xcode 10.1 Build version 10B61
```

Plugins:
```
cl.rmd.cordova.dialoggps 0.0.2 "DialogGPS"
com.lampa.startapp 6.1.6 "startApp"
com.ludei.webview.plus 2.3.0 "Webview+"
com.peerio.cordova.plugin.carrier 1.0.1 "Carrier"
com.phonegap.plugins.nativesettingsopener 1.0.1 "Native settings"
cordova-admobsdk 7.37.0 "Google Mobile Ads SDK for Cordova"
cordova-custom-config 5.1.0 "cordova-custom-config"
cordova-instagram-plugin 0.5.8 "Instagram"
cordova-launch-review 3.1.1 "Launch Review"
[...]

PR #4162, opened 2019-09-21
Additional #4156 work

brownoxford:
Some additional updates related to #4156 to ensure that stdio is properly set for 'runServeDeploy()' and 'runBuildDeploy()' so that `ionic cordova run` functions as expected in all code paths, specifically with various combinations of `--livereload`, `--verbose`, and `--no-native-run`.

fixes https://github.com/ionic-team/ionic-cli/issues/4165

### DIFF
--- a/packages/ionic/src/commands/cordova/run.ts
+++ b/packages/ionic/src/commands/cordova/run.ts
@@ -334,14 +334,19 @@ Just like with ${input('ionic cordova build')}, you can pass additional options 
       }
     }
 
+    const buildOpts: IShellRunOptions = { };
+    // ignore very verbose compiler output unless --verbose (still pipe stderr)
+    if (!options['verbose']) {
+      buildOpts.stdio = ['ignore', 'ignore', 'pipe'];
+    }
+
     if (options['native-run']) {
       const [ platform ] = inputs;
       const packagePath = await getPackagePath(conf.getProjectInfo().name, platform, { emulator: !options['device'], release: !!options['release'] });
-
-      await this.runCordova(filterArgumentsForCordova({ ...metadata, name: 'build' }, options), { stdio: 'inherit' });
+      await this.runCordova(filterArgumentsForCordova({ ...metadata, name: 'build' }, options), buildOpts);
       await this.runNativeRun(createNativeRunArgs({ packagePath, platform }, { ...options, connect: false }));
     } else {
-      await this.runCordova(filterArgumentsForCordova(metadata, options), { stdio: 'inherit' });
+      await this.runCordova(filterArgumentsForCordova(metadata, options), buildOpts);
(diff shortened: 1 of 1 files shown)


packages/ionic/src/commands/cordova/run.ts
```diff
@@ -334,14 +331,17 @@ Just like with ${input('ionic cordova build')}, you can pass additional options
       }
     }
 
+    const buildOpts: IShellRunOptions = { };
+    // ignore very verbose compiler output on stdout unless --verbose
+    buildOpts.stdio = options['verbose'] ? 'inherit' : ['pipe', 'ignore', 'pipe'];
```
dwieeb: For builds we should show all output always.

dwieeb:
@brownoxford Would this fix this issue: https://github.com/ionic-team/ionic-cli/issues/4165?

packages/ionic/src/commands/cordova/run.ts
brownoxford: Ah, ok, I was duplicating what I saw in runServeDeploy for the handling of output - I'll remove all changes to runBuildDeploy as it was otherwise working wrt. #4165

brownoxford:
> @brownoxford Would this fix this issue: #4165?

The changes to runServeDeploy fix #4165 in the case where the user specified --no-native-run, as stdio was only being addressed when using native run, and then was only piping stderr (stdin also needed to be piped).

brownoxford:
Sorry, confused 4156 with 4165 reading too quickly - reviewing now.

brownoxford:
Ok, No, I don't believe this fixes #4165 as that user reported using native-run and the only thing related that this addresses is a --no-native-run code path. The user in #4165 needs to update ionic CLI to get your fixes for #4156.

brownoxford:
Ok, I've reverted changes to runBuildDeploy

Yes, this does fix #4165 by changing the stdio options from `['ignore', 'ignore', 'pipe']` to `['pipe', 'ignore', 'pipe']`

dwieeb:
Thanks!! This is great! 💙

dwieeb:
CLI 5.4.2 has this fix!

bnallmaster:
I updated to 5.4.2 and am still seeing the iOS livereload hang up issue. I also updated my native-run to the latest.

hijoncon:
No luck even on the updated catalina...

dwieeb:
I'm assuming the iOS hanging issue everyone is still having is [this issue](https://github.com/ionic-team/native-run/issues/80), which is unrelated to the fixes in this PR.

If the command is hanging during the `native-run` option, try this: https://github.com/ionic-team/ionic-cli/issues/4165#issuecomment-539530548

Issue #4196, opened 2019-10-30
ionic repair fails if package-lock.json don't exists.


ludufre:
**Description:**
`ionic repair` fails if package-lock.json don't exists.

**Steps to Reproduce:**
- Start a new Ionic project.
- Do `npm install`
- Remove `package-lock.json`
- Run `ionic repair`

**Output:**
```zsh
ludufre@Mac-mini-de-Luan sonar % ionic repair
[INFO] ionic repair will do the following:
       
       - Remove node_modules/ and package-lock.json
       - Run npm i to restore dependencies
       - Remove platforms/ and plugins/
       - Run cordova prepare to restore platforms and plugins
       
? Continue? Yes

Error: ENOENT: no such file or directory, unlink '/Users/ludufre/Desktop/REPOs/sonar/package-lock.json'
⠧ Removing ./package-lock.json
```
(Forever with message "Removind...")

Solution: `touch package-lock.json` and then run `ionic repair` again.

**My `ionic info`:**
```
Ionic:

   Ionic CLI                     : 5.4.4 (/Users/ludufre/.npm-packages/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.11.2
   @angular-devkit/build-angular : 0.803.15
   @angular-devkit/schematics    : 8.3.15
   @angular/cli                  : 8.3.15
   @ionic/angular-toolkit        : 2.1.1

Cordova:

   Cordova CLI       : 9.0.0 (cordova-lib@9.0.1)
   Cordova Platforms : android 8.1.0, browser 6.0.0, ios 5.0.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 4.1.2, (and 22 other plugins)

Utility:

   cordova-res : 0.8.1 
   native-run  : not installed

System:

   NodeJS : v12.13.0 (/usr/local/bin/node)
   npm    : 6.12.0
   OS     : macOS Catalina
```

**Other Information:**
Solution: `touch package-lock.json` and then run `ionic repair` again.

dwieeb:
The fix by @NightmareZ was released in 5.4.5. Thanks for the report!

PR #4197, opened 2019-10-30
fix(cli): ionic repair fails if package-lock.json don't exists. (#4196)

NightmareZ:


dwieeb:
Thanks! I'll release this now.

PR #4254, opened 2019-12-17
feat: `ionic capacitor build` command

dwieeb:
I created this PR to spark discussion. Right now, there is no functional difference between this new `ionic cap build` command the existing `ionic cap run` command without `--livereload`. This is because each command does the following:
1. Perform ionic build
2. Copy web assets into the specified native platform
3. Open the IDE for your native project (Xcode for iOS, Android Studio for Android)

The differentiation between "running your app" and "building your project" happens outside of the Ionic CLI. It's up to the dev to click build or run within the Native IDE. This may change if Capacitor gets programmatic build/run functionality, in which case these two commands would differ.


jcesarmobile:
Capacitor copy with no platform copies to all installed platforms, maybe we can do that instead of prompting if no platform is specified.
Since Capacitor supports electron, this PR could also include electron as possible platform, but for angular electron requires to pass —base-href=./ (at least for angular, not sure about the other frameworks)

dwieeb:
> Capacitor copy with no platform copies to all installed platforms, maybe we can do that instead of prompting if no platform is specified.

I suppose we could have this command build for all platforms, but that behavior would differ from `ionic cordova build`, which largely inspires this command. It would also be different from `ionic cap run`.

> Since Capacitor supports electron, this PR could also include electron as possible platform

Maybe, but that would mean Ionic would have to support it. I think for now we want to stick to iOS and Android.

bryplano:
So just to be clear, for an Electron app, could I still do `ionic capacitor build electron` and have the first two steps from above work? Or we're ruling out Electron entirely?

I agree with @jcesarmobile, it would be great if we can give Electron a little love *just* for Capacitor applications. An Electron + Capacitor app is actually what inspired me to look into a command like this, because it was a pain to remember to `ionic build` + `npx cap copy` all the time.

Yes, you can combine them with a script, but why not make it easy? :D

kensodemann:
I think @dwieeb makes a _realy_ good point here though:

> Maybe, but that would mean Ionic would have to support it. I think for now we want to stick to iOS and Android.

IIR, Electron support is not something that is fully fleshed out yet, and we should probably hold off on adding it here until it is something that is better fleshed out (this is all out of recollection, though, mind you, and things may have changed since my last discussions on this)

dwieeb:
One difference between this and `ionic cap run` is that `ionic cap build` should probably run `cap sync` instead of just `cap copy`. What do you think @jcesarmobile?

jcesarmobile:
Makes sense

Issue #4114, opened 2019-08-09
[@ionic/v1-toolkit] Issue with outdated gulp dependency


CheetahDev:
**Description:**
As part of a general development environment update, we decided to update some of our dev dependencies including gulp and gulp-* utilities. Our projects use Ionic v.1.3.5 with up-to-date Ionic CLI and @ionic/v1-toolkit. We updated our gulpfile.js file according to the latest version of gulp (we were using gulp@3.9.1). Once the code has been updated, we used the normal `ionic serve` command to start the local server. The server started up but no tasks have been executed. We run the command again but with the --verbose option. This highlighted the following issue:
```
[v1] 2019-08-09T14:37:54.498Z ionic:v1-toolkit:lib:gulp Using gulpfile: /Users/xxxx/dev/NEXT-MASTER-new-gulp/project/gulpfile.js
[v1] 2019-08-09T14:37:54.499Z ionic:v1-toolkit:lib:gulp Using gulp: /Users/xxxx/dev/NEXT-MASTER-new-gulp/project/node_modules/@ionic/v1-toolkit/node_modules/gulp/index.js
[v1] 2019-08-09T14:37:55.875Z ionic:v1-toolkit:lib:gulp Loaded gulp tasks: {}
[v1] [16:37:55] Serving directory www
```
I kept the previous version aside and ran the same command:
```
[v1] 2019-08-09T14:28:43.114Z ionic:v1-toolkit:lib:gulp Using gulpfile: /Users/xxxx/dev/NEXT-MASTER-old-gulp/project/gulpfile.js
[v1] 2019-08-09T14:28:43.115Z ionic:v1-toolkit:lib:gulp Using gulp: /Users/xxxx/dev/NEXT-MASTER-old-gulp/project/node_modules/gulp/index.js


[...]

PR #4264, opened 2019-12-27
@ionic/v1-toolkit: Add compatibility with gulp v4 file

blavenie:
- Before executing the 'ionic:serve:before' task, detect if `gulpfile.js` has some exported function, as a standard gulp v4 file.
- If v4 file is detected, then define all exported fucntions as gulv v3 task using `gulp.task(<name>, [], exportedFunction)`

This PR should fix the issue #4114

Example of `gulpfile.js` : 
```javascript
const gulp = require('gulp'),
  sass = require('gulp-sass');

function sassTask(done) {
  return gulp.src('./scss/ionic.app.scss')
    .pipe(sass()).on('error', sass.logError)
    .pipe(gulp.dest('./www/css/'))
    .on('end', done); 
}

function watchTask(done) {
    gulp.watch('./scss/ionic.app.scss', sassTask);
    done();
}

(...)
// Define exported task (in the gulp v4 way)
exports.sass = executeSass;
exports['ionic:serve:before'] = gulp.series(sassTask, watchTask);

```
Example of `package.json` : 
```json
devDependencies : {
  "gulp": "4.0.2",
  "gulp-sass": "^4.0.2",
  "node-sass": "^4.13.0"
   (...)
}
```


dwieeb:
This looks good, but will need some cleanup so the build process finishes

blavenie:
> will need some cleanup so the build process finishes

Can you explain to me ? What do you expected from me ?

dwieeb:
The CI build is failing due to TS compilation errors. They will need to be fixed before this is merged. Thanks!

asa47:
Hello @blavenie and @dwieeb,
Any update on this issue? Or any advice on how to fix it?

blavenie:
ok @dwieeb, all JS error seems to be fixed :)

dwieeb:
Awesome! Thanks!!

dwieeb:
`@ionic/v1-toolkit@3.2.0` is released with this feature! Great work. 🍻

rodrigograca31:
In my opinion, this still needs more work.

If you still support an ionic v1 app nowadays with the node version 12 we have to use gulp 4

You can update your `gulpfile.js` and gulp version to 4 and then run `gulp default` or whatever your task is but when you run `ionic serve` it will run gulp 3.9 even if we installed gulp 4 because `@ionic/v1-toolkit` depends on it.

Actually it runs other gulp stuff and then it checks for tasks called `ionic:serve:before` and etc and runs them
```
> ionic-v1 serve --host=localhost --port=8100 --livereload-port=35729 --dev-port=53703 --engine=browser
[v1] 2020-07-29T11:52:24.764Z ionic:v1-toolkit:lib:gulp Using gulpfile: /home/USER/git/PROJECT/gulpfile.js
[v1] 2020-07-29T11:52:24.765Z ionic:v1-toolkit:lib:gulp Using gulp: /home/USER/git/PROJECT/node_modules/@ionic/v1-toolkit/node_modules/gulp/index.js
[v1] [12:52:24] Serving directory www
[v1] [12:52:24] Cannot load gulp: ReferenceError: primordials is not defined
```

Upon further inspection it breaks here:
because it requires gulp and its 3.9 with node 12 and that explodes 💥
```
try {
            const gulpPath = require.resolve('gulp');
            debug(`Using gulp: ${gulpPath}`);
            _gulpInst = require(gulpPath);
        }
        catch (e) {
            if (e.code !== 'MODULE_NOT_FOUND') {
                throw e;
            }
            throw new Error(chalk.red(`Cannot find module 'gulp'`));
        }
```

Related: #4114 

There should be a way to either use the globally installed gulp or a locally installed one on the main `package.json` instead of the one that comes with `@ionic/v1-toolkit`

These are the critical lines:
```
const gulpPath = require.resolve('gulp');
            debug(`Using gulp: ${gulpPath}`);
            _gulpInst = require(gulpPath);
```

Issue #4192, opened 2019-10-28
Angular*CLI.buildArchitectCommand should not fallback to app


SchnWalter:
The [`AngularBuildCLI.buildArchitectCommand()`](https://github.com/ionic-team/ionic-cli/blob/ionic%405.4.4/packages/ionic/src/lib/project/angular/build.ts#L164) and [`AngularServeCLI.buildArchitectCommand()`](https://github.com/ionic-team/ionic-cli/blob/ionic%405.4.4/packages/ionic/src/lib/project/angular/serve.ts#L268) methods should not hardcode `app` as a fallback project name when no project name is provided via arguments. The best approach would be to read the `defaultProject` key from `angular.json` or by running `ng config defaultProject`

With this, the Ionic CLI will be able to work, by default, in Angular workspaces where the `defaultProject` is not named `app`; as for the ones generated using the `ng add @ionic/angular` schematic; see: #4191

L.E. #4121 should be dealt with first.

dwieeb:
@SchnWalter How can I reproduce the error case where it's falling back to `app` in a multi-app environment? In my test repo, the Ionic CLI properly warns about not knowing which project to use for `ionic build`, `ionic serve`, and `ionic g`.

<img width="849" alt="image" src="https://user-images.githubusercontent.com/236501/73809536-6b53cd80-4788-11ea-93e1-9ae298d85251.png">

The "bare projects" error isn't the ideal message, but it seems to error properly when it's supposed to. If I try to manipulate `ionic.config.json` in a way where it treats the project in a way that's both single-app and multi-app, I get this correct error:

<img width="824" alt="image" src="https://user-images.githubusercontent.com/236501/73809786-21b7b280-4789-11ea-9dd8-456f50651e89.png">

See my test repo branches:

* https://github.com/dwieeb/ionic-multiapp/tree/example1 (no root app, hoisted deps)
* https://github.com/dwieeb/ionic-multiapp/tree/example2 (root app + second app in subdir)

What am I missing?

SchnWalter:
You need to use Angular CLI to generate a new workspace with at least an ionic application and some other angular specific library or application. Somewhat related to the fix from here https://github.com/ionic-team/ionic/pull/19765, which was merged but it had a regression that was fixed in https://github.com/ionic-team/ionic/pull/20281.

To reproduce this step you can checkout https://github.com/SchnWalter/ionic-cli-issue-4192 (or follow the "Steps for recreating the demo repository" section below.

[...]

PR #4311, opened 2020-02-04
feat(angular): read the default project from workspace configuration

SchnWalter:
When the notion of workspaces was introduced in Angular, the Ionic CLI was updated with a hardcoded "app" as the defaultProject fallback, which is the default project name for newly generated Ionic+Angular projects. In this MR I've converted all build, generate and serve runner instances to read options using promises and updated the angular runners to read the default project form the Angular workspace configuration (the root `angular.json` file).

In case where you worked hard to make the Ionic Angular application in a multi-app Angular workspace, the only working solution was to always provide the Ionic project name in all CLI commands, so this isn't really a breaking change. There are some other edge cases where this might break your build scripts, but if you worked that hard to make them work in those situations, probably you would have realized that the generate command actually had a bug (which this MR also fixes and tests were added), so you would have change the scripts to always provide the project name. So, no breaking change, per-se in this MR.

Fixes https://github.com/ionic-team/ionic-cli/issues/4192


packages/@ionic/cli/src/lib/project/angular/index.ts
```diff
@@ -79,6 +91,20 @@ export class AngularProject extends Project {
...
+      const angularJsonPath = resolve(this.directory, 'angular.json');
+      const angularWorkspace: WorkspaceSchema = await readJson(angularJsonPath);
+      if (angularWorkspace && angularWorkspace.defaultProject && angularWorkspace.defaultProject !== '') {
+        return Promise.resolve(angularWorkspace.defaultProject);
+      }
+    } catch (e) {
+      // ignore
+    }
+
+    return Promise.resolve('app');
```
SchnWalter: Note this MR still keeps the hardcoded default fallback of `app`, in the angular world, this should have actually been the first defined project, but this would break build scripts for people that rely on this fallback. Maybe this can be marked as a deprecated "feature" and a future major release could use the proper fallback.

packages/@ionic/cli/src/lib/project/angular/generate.ts
```diff
@@ -98,14 +98,15 @@ To test a generator before file modifications are made, use the ${input('--dry-r
...
+  async createOptionsFromCommandLine(inputs: CommandLineInputs, options: CommandLineOptions): Promise<AngularGenerateOptions> {
+    const baseOptions: GenerateOptions = await super.createOptionsFromCommandLine(inputs, options);
+    const project = options['project'] ? String(options['project']) : await this.e.project.getWorkspaceDefaultProject();
 
     // TODO: this is a little gross, is there a better way to pass in all the
     // options that the command got?
     return {
       ...lodash.omit(options, '_', '--', ...GLOBAL_OPTIONS.map(opt => opt.name)),
-      project: options['project'],
+      project,
```
SchnWalter: The `ionic generate` angular command was the only one that didn't provide the default project name fallback (`app`). I've added it here and added a few tests.

SchnWalter:
Closing in favor of: https://github.com/ionic-team/ionic-cli/commit/42c51996632cdbc6ad6e514f36a584e3a450e380

Issue #4285, opened 2020-01-16
[Feature request] Add support for pnpm


duong755:
Hi,
I wonder will ionic-cli support the third `npmClient` - pnpm?

dwieeb:
I am happy to look at pull requests that add pnpm to the `pkgManagerArgs` function: https://github.com/ionic-team/ionic-cli/blob/develop/packages/ionic/src/lib/utils/npm.ts#L44 

But, as of now this is low priority for us. Thanks for understanding!

PR #4330, opened 2020-02-21
feat: add support for pnpm

muuvmuuv:
This PR adds support for the package manager `pnpm`.

More info about `pnpm `: https://pnpm.js.org/en/

Closes: #4285

### DIFF
--- a/packages/@ionic/cli/src/definitions.ts
+++ b/packages/@ionic/cli/src/definitions.ts
@@ -392,7 +392,7 @@ export interface ITelemetry {
   sendCommand(command: string, args: string[]): Promise<void>;
 }
 
-export type NpmClient = 'yarn' | 'npm';
+export type NpmClient = 'yarn' | 'npm' | 'pnpm';
 
 export type FeatureId = 'ssl-commands';
 
--- a/packages/@ionic/cli/src/lib/utils/__tests__/npm.ts
+++ b/packages/@ionic/cli/src/lib/utils/__tests__/npm.ts
@@ -211,7 +211,7 @@ describe('@ionic/cli', () => {
 
         it('should be run args with script and script args', async () => {
           const result = await pkgManagerArgs('pnpm', { command: 'run', script: 'test', scriptArgs: ['-s'] });
-          expect(result).toEqual(['pnpm', 'run',  'test', '-s']);
+          expect(result).toEqual(['pnpm', 'run',  'test', '--', '-s']);
         });
 
         it('should be info args', async () => {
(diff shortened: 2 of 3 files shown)


dwieeb:
@muuvmuuv This is great!

One question: it seems like pnpm need the `--` arg separator (like in npm) for `pnpm run` (see https://pnpm.js.org/en/cli/run#synopsis)

It looks like it's not being added: https://github.com/ionic-team/ionic-cli/pull/4330/files#diff-1b40833217c4862ef2a34651e286cc04R214

muuvmuuv:
@dwieeb oh yes! Sorry. I will fix it probably next week but we are celebrating carnival this weeks so it might be delayed

dwieeb:
Awesome work!!

aparajita:
While this is appreciated, it doesn't quite go all the way. `ionic build` fails because of this line:

https://github.com/ionic-team/ionic-cli/blob/483d9ab593c4715ab21b3b721c6735c7e01d367e/packages/%40ionic/cli/src/lib/build.ts#L47

muuvmuuv:
Any idea how I could test this @aparajita ? I think the best is to have a test that build a sample ionic ng/vue/etc. project. cc: @dwieeb

aparajita:
@muuvmuuv In any ionic project, set `npmClient=pnpm` in `ionic.config.js`, run `ionic build`, and it will fail.

muuvmuuv:
No, I mean with Jest @aparajita so future added pm's (if there will be any)  will not fail or updated API's

aparajita:
Here are the conditions under which it fails:

In `package.json`, assuming you have a "build" script:

```json
"scripts": {
  "ionic:build": "pnpm run build"
}
```

Then in a terminal:

```sh
ionic config set npmClient pnpm
ionic config set type custom
ionic build
```

You get this output:

```
> yarn run --non-interactive ionic:build

[ERROR] yarn is required for this command to work properly.
```

aparajita:
I should point out I don't have yarn installed and never will. 😁

aparajita:
@muuvmuuv I have a working fix, I'll make a PR later today.

muuvmuuv:
Great, set me on CC I would love to have a look at it. I also have worked on a fix but couldn't test it.

aparajita:
I have **ionic** fixed, but **capacitor** still needs fixing unfortunately. And it's a much bigger job.

aparajita:
I'm quite surprised that the authors of capacitor wrote the code in a binary way: if you're not using yarn, you must be using npm. As if the universe of package managers would eternally remain the same! 🙄

aparajita:
Hey @muuvmuuv, the PR is here: https://github.com/ionic-team/ionic-cli/pull/4459

PR #4345, opened 2020-03-06
Start wizard

mlynch:
This pulls down the app created on the website, initializes all the values for the app start, updates a theme file w/ a theme color, and writes it back.

Currently looking for a review in terms of whether I've put the functionality in the right place or anything else that I should fix. 

Also I'll be fixing the `START_WIZARD_URL` env var later

### DIFF
--- a/packages/@ionic/cli/src/commands/start.ts
+++ b/packages/@ionic/cli/src/commands/start.ts
@@ -27,10 +27,11 @@ interface StartWizardApp {
   template: string;
   'package-id': string;
   tid: string;
-  atk: string;
   email: string;
   cid: number;
   theme: string;
+  ip: string;
+  utm: { [key:string]: string }
 }
 
 export class StartCommand extends Command implements CommandPreRun {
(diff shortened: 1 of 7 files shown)


packages/@ionic/cli/src/commands/start.ts
```diff
@@ -129,10 +143,51 @@ Use the ${input('--type')} option to start projects using older versions of Ioni
...
+  async startIdStart(inputs: CommandLineInputs, options: CommandLineOptions) {
+    const startId = options['start-id'];
+
+    const wizardApiUrl = process.env.START_WIZARD_URL;
+
+    if (!wizardApiUrl) {
+      return;
+    }
+
+    const { req } = await createRequest('GET', `${wizardApiUrl}/api/v1/wizard/app/${startId}`, this.env.config.getHTTPConfig());
```
dwieeb: `this.env.client` usage is preferred

packages/@ionic/cli/src/lib/project/index.ts
```diff
@@ -563,11 +564,66 @@ export abstract class Project implements IProject {
...
+        themePath = path.join(this.directory, 'src', 'theme', 'variables.css');
+        break;
+      }
+    }
+    if (!themePath) {
+      return;
+    }
+
+    return this.writeThemeColor(themePath, themeColor);
+  }
```
dwieeb: This likely belongs in the `project/angular` Project class, as it should only run for Angular projects.

Issue #4374, opened 2020-03-23
native-run searching for unsigned apk while apk is already signed


wellwoller:
**Version:**

0.3.0

**Description:**

This happened after I upgraded to ionic 5, angular 9 etc. (Basically a full upgrade of my npm modules, ionic modules and cordova + plugins)
While running `ionic cordova run android -- device --prod --release` I get an error because native-run can't find the unsigned version of my apk. As you can see below my app is being built and signed afterwards.  But native-run THEN tries to look for the unsigned apk in that same directory which obviously can't be found at that moment. Debug build and even prod build are running fine but the release build won't push to device and while I can manually install the apk to my device there seems to be issues. 

**Command Output:**

`BUILD SUCCESSFUL in 20s
43 actionable tasks: 41 executed, 2 up-to-date

C:\Users\{User}\Repos\{MyApp}>if "0" == "0" goto mainEnd

C:\Users\{User}\Repos\{MyApp}>if "Windows_NT" == "Windows_NT" endlocal
Command finished with error code 0: C:\Users\{User}\Repos\{MyApp}\platforms\android\gradlew cdvBuildRelease,-b,C:\Users\{User}\Repos\{MyApp}\platforms\android\build.gradle
Built the following apk(s): 
        C:\Users\{User}\Repos\{MyApp}\platforms\android\app\build\outputs\apk\release\app-release.apk
No scripts found for hook "after_compile".
No scripts found for hook "after_build".
> native-run.cmd android --app platforms\android\app\build\outputs\apk\release\app-release-unsigned.apk --device --verbose
[native-run] 2020-03-23T14:46:40.272Z native-run:android:utils:sdk:resolveSDKRoot Looking for $ANDROID_HOME
[native-run] 2020-03-23T14:46:40.277Z native-run:android:utils:sdk:resolveSDKRoot Using $ANDROID_HOME at C:\Users\{User}\AppData\Local\Android\Sdk
[native-run] 2020-03-23T14:46:40.277Z native-run:android:utils:sdk:resolveEmulatorHome Looking for $ANDROID_EMULATOR_HOME
[native-run] 2020-03-23T14:46:40.277Z native-run:android:utils:sdk:resolveEmulatorHome Looking at $HOME/.android
[native-run] 2020-03-23T14:46:40.277Z native-run:android:utils:sdk:resolveEmulatorHome Using $HOME/.android/ at C:\Users\{User}\.android
[...]

PR #4377, opened 2020-04-02
fix(cordova): use output.json to deploy correct apk

dwieeb:
fixes https://github.com/ionic-team/ionic-cli/issues/4374


dwieeb:
😂

PR #4403, opened 2020-04-25
Added the --no-open option

Heerschop:
Added the option `--no-open` Do not invoke Capacitor open, to the ionic capacitor `build` and `run` commands.

This option will make it easier to use ionic capacitor from the command line, especially when you what to use the `--livereload` option.

Currently, if you what to use live reload from the CLI. You can not use the `ionic capacitor run <platform> --livereload` command, because it will always start the corresponding IDE. So you have to write custom scripts that will update the `capacitor.config.json` file with the correct information needed for live reload to work. This pull request can make this use case simpler.


dwieeb:
Love it! Great work. 🍻

dwieeb:
6.7.0 is out with this feature!

PR #4417, opened 2020-05-03
Ionic capacitor hooks

Heerschop:
This pull request contains changes to simplify usage of `ionic capacitor` from the CLI and in a CI/CD environment. By adding two extra hooks and by adding environment variables for all the hook shell scripts.

Added four ionic capacitor CLI [Hooks](https://ionicframework.com/docs/cli/configuration#hooks)

- `ionic:capacitor:run:before` : executed on run before capacitor open is executed
- `ionic:capacitor:run:after` : executed on run after capacitor open is executed
- `ionic:capacitor:build:before` : executed on build before capacitor open is executed
- `ionic:capacitor:build:after` : executed on build after capacitor open is executed

The capacitor hooks are executed after the `ionic:serve` and `ionic:build` hooks.

Updated the task run method. The task run method will now supply the shell scripts with generated environment variables based on the information in the Hook Input object.

# Example usage
#### Content example script: build-before.sh
``` sh
#!/bin/bash
echo Hook build-before.sh: $0 $*
env | grep IONIC_CLI_HOOK_CTX
```
#### Content example script: capacitor-build-before.sh
``` sh
#!/bin/bash
echo Hook capacitor-build-before.sh: $0 $*
env | grep IONIC_CLI_HOOK_CTX
```
#### Content: package.json
``` text
{
  ...
  "scripts": {
    ...
    "ionic:build:before:after": "scripts/build-before.sh",
    "ionic:capacitor:build:before": "scripts/capacitor-build-before.sh"
  },
  ...
}
```

> test-cli-app@0.0.1 ionic:build:before /home/bas/Documents/temp/test-cli-app
> scripts/build-before.sh

``` bash
Hook build-before.sh: scripts/build-before.sh
IONIC_CLI_HOOK_CTX_NAME=build:before
IONIC_CLI_HOOK_CTX_BUILD_CORDOVA_ASSETS=true
IONIC_CLI_HOOK_CTX_BUILD_ENGINE=capacitor
IONIC_CLI_HOOK_CTX_BUILD_PLATFORM=android
IONIC_CLI_HOOK_CTX_BUILD_PROJECT=app
IONIC_CLI_HOOK_CTX_BUILD_TYPE=angular
IONIC_CLI_HOOK_CTX_BUILD_VERBOSE=false
```
> ng run app:build
Generating ES5 bundles for differential loading...
ES5 bundle generation complete.

chunk {polyfills} polyfills-es2015.js, polyfills-es2015.js.map (polyfills) 297 kB [initial] [rendered]
....................
chunk {vendor} vendor-es5.js, vendor-es5.js.map (vendor) 5.39 MB [initial] [rendered]
Date: 2020-05-03T14:19:04.541Z - Hash: cd8ac09eeb78f0a5144a - Time: 9751ms
> capacitor sync android
✔ Copying web assets from www to android/app/src/main/assets/public in 385.94ms
✔ Copying native bridge in 408.54μp
✔ Copying capacitor.config.json in 624.42μp
✔ copy in 397.01ms
✔ Updating Android plugins in 3.13ms
  Found 0 Capacitor plugins for android:
✔ update android in 8.56ms
Sync finished in 0.414s
> npm run ionic:capacitor:build:before

> test-cli-app@0.0.1 ionic:capacitor:build /home/bas/Documents/temp/test-cli-app
> scripts/capacitor-build-before.sh

``` bash
Hook capacitor-build.sh: scripts/capacitor-build-before.sh
IONIC_CLI_HOOK_CTX_NAME=capacitor:build:before
IONIC_CLI_HOOK_CTX_BUILD_CORDOVA_ASSETS=true
IONIC_CLI_HOOK_CTX_BUILD_ENGINE=browser
IONIC_CLI_HOOK_CTX_BUILD_PROJECT=app
IONIC_CLI_HOOK_CTX_BUILD_TYPE=angular
IONIC_CLI_HOOK_CTX_BUILD_VERBOSE=false
IONIC_CLI_HOOK_CTX_CAPACITOR_APP_ID=io.ionic.starter
IONIC_CLI_HOOK_CTX_CAPACITOR_APP_NAME=test-cli-app
IONIC_CLI_HOOK_CTX_CAPACITOR_VERBOSE=false
```

[INFO] Ready for use in your Native IDE!
       
       To continue, build your project using Android Studio!

> capacitor open android


packages/@ionic/cli/src/definitions.ts
```diff
@@ -576,6 +586,14 @@ export interface VueBuildOptions extends BuildOptions<'vue'> {
   cordovaAssets?: boolean;
 }
 
+export interface IonicCapacitorOptions {
+  configuration?: string;
```
dwieeb, line 590: `configuration` is specific to Angular. We should reuse the hook input from build/serve, which is the same as the runner options. https://github.com/ionic-team/ionic-cli/blob/f4f05ffe4649a66f4c7961ee09bea97343d0e4f3/packages/%40ionic/cli/src/lib/build.ts#L93-L95

https://github.com/ionic-team/ionic-cli/blob/f4f05ffe4649a66f4c7961ee09bea97343d0e4f3/packages/@ionic/cli/src/commands/capacitor/base.ts#L102

packages/@ionic/cli/src/lib/hooks.ts
```diff
@@ -97,6 +104,23 @@ export abstract class Hook {
...
+    let environment: NodeJS.ProcessEnv = {};
+
+    for (const [key, value] of Object.entries(input)) {
+      if (typeof value === 'object') {
+        environment = {
+          ...environment,
+          ...this.generateCTXEnvironment(value),
+        };
+      } else {
+        environment[ 'IONIC_CLI_HOOK_CTX_' + key.toUpperCase() ] = value;
```
dwieeb, line 118: `key` might have weird characters, I suggest this:

```suggestion
        environment[`IONIC_CLI_HOOK_CTX_${lodash.snakeCase(key.toUpperCase())}`] = value;
```

packages/@ionic/cli/src/commands/capacitor/build.ts
```diff
@@ -142,4 +151,22 @@ To configure your native project, see the common configuration docs[^capacitor-n
...
+        throw new FatalException(e.message);
+      }
+
+      throw e;
+    }
+  }
+}
+
+class CapacitorRunHook extends Hook {
+  readonly name = 'capacitor:build';
```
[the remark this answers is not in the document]
Heerschop: I think the value `capacitor:build` of the name field is correct, but we should rename the class to `CapacitorBuildHook`

packages/@ionic/cli/src/lib/hooks.ts
Heerschop: That's definitely better

packages/@ionic/cli/src/definitions.ts
Heerschop: If we do that, then we'll rule out adding other capacitor-specific information in  later update, like the `appId` for instance.

packages/@ionic/cli/src/commands/capacitor/build.ts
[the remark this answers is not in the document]
dwieeb: ah yep, whoops

packages/@ionic/cli/src/definitions.ts
dwieeb: I was thinking `CapacitorBuildHookInput` would have a `build` key much like `BuildHookInput` and then have an additional key of Capacitor-specific info. `CapacitorRunHookInput` would have either a `build` or a `serve` key, depending on if `--livereload` is used or not. Does that make sense?

packages/@ionic/cli/src/commands/capacitor/build.ts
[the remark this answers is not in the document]
Heerschop: :-) i should have used the correct class name in the first place.

packages/@ionic/cli/src/definitions.ts
Heerschop: That sounds great, in that way it is even more flexible.

packages/@ionic/cli/src/commands/capacitor/build.ts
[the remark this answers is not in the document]
Heerschop: Renamed class to `CapacitorBuildHook`

packages/@ionic/cli/src/lib/hooks.ts
Heerschop: Changed it to
``` javascript
const name = [path, key].join('_');

environment[`IONIC_CLI_HOOK_CTX_${lodash.snakeCase(name)}`.toUpperCase()] = value;
```

Output now looks like:
``` bash
IONIC_CLI_HOOK_CTX_NAME=capacitor:run
IONIC_CLI_HOOK_CTX_CAPACITOR_APP_ID=io.ionic.starter
IONIC_CLI_HOOK_CTX_CAPACITOR_APP_NAME=test-cli-app
IONIC_CLI_HOOK_CTX_CAPACITOR_SERVER_URL=http://localhost:8100
IONIC_CLI_HOOK_CTX_CAPACITOR_VERBOSE=false
IONIC_CLI_HOOK_CTX_SERVE_ENGINE=browser
IONIC_CLI_HOOK_CTX_SERVE_EXTERNAL_ADDRESS_REQUIRED=false
IONIC_CLI_HOOK_CTX_SERVE_HOST=localhost
IONIC_CLI_HOOK_CTX_SERVE_LAB=false
IONIC_CLI_HOOK_CTX_SERVE_LAB_HOST=localhost
IONIC_CLI_HOOK_CTX_SERVE_LAB_PORT=8200
IONIC_CLI_HOOK_CTX_SERVE_LIVERELOAD=true
IONIC_CLI_HOOK_CTX_SERVE_OPEN=false
IONIC_CLI_HOOK_CTX_SERVE_PORT=8100
IONIC_CLI_HOOK_CTX_SERVE_PROJECT=app
IONIC_CLI_HOOK_CTX_SERVE_PROXY=true
IONIC_CLI_HOOK_CTX_SERVE_VERBOSE=false
```

packages/@ionic/cli/src/definitions.ts
Heerschop: I have updated the pull request, the hook input now looks like this:

``` typescript
export interface CapacitorRunHookInput {
  readonly name: 'capacitor:run';
  readonly serve?: AngularServeOptions | IonicAngularServeOptions | Ionic1ServeOptions;
  readonly build?: AngularBuildOptions | IonicAngularBuildOptions | Ionic1BuildOptions;
  readonly capacitor: IonicCapacitorOptions;
}

export interface CapacitorBuildHookInput {
  readonly name: 'capacitor:build';
  readonly build: AngularBuildOptions | IonicAngularBuildOptions | Ionic1BuildOptions;
  readonly capacitor: IonicCapacitorOptions;
}

export interface IonicCapacitorOptions {
  verbose?: boolean;
  appId?: string;
  appName?: string;
  server?: {
    url?: string;
  };
  '--': string[];
}
```

Heerschop:
I'm not sure why these test are failing, the current develop branch is failing as well. I'm pretty sure it has nothing to do with the changes I have made

packages/@ionic/cli/src/commands/capacitor/run.ts
```diff
@@ -222,4 +229,39 @@ For Android and iOS, you can setup Remote Debugging on your device with browser
       `To continue, run your project on a device or ${getVirtualDeviceNameForPlatform(platform)} using ${getNativeIDEForPlatform(platform)}!`
     );
   }
+
+  private async runCapacitorHook(project: IProject , inputs: CommandLineInputs, options: CommandLineOptions, e: HookDeps): Promise<void> {
+    const hook = new CapacitorRunHook(e);
+    let serveOptions: ServeOptions | undefined;
+    let buildOptions: any | undefined;
```
dwieeb, line 236: Maybe now is the time we create a union of the different BuildOptions and ServeOptions so we can reuse them.

In `definitions.ts` we should define maybe something like this:

```ts
export type AnyBuildOptions = ReactBuildOptions | AngularBuildOptions | IonicAngularBuildOptions | Ionic1BuildOptions;
export type AnyServeOptions = ReactServeOptions | AngularServeOptions | IonicAngularServeOptions | Ionic1ServeOptions;
```

Then you can use them in your hooks:

```ts
export interface CapacitorRunHookInput {
  readonly name: 'capacitor:run';
  readonly serve?: AnyServeOptions;
  readonly build?: AnyBuildOptions;
  readonly capacitor: IonicCapacitorOptions;
}

export interface CapacitorBuildHookInput {
  readonly name: 'capacitor:build';
  readonly build: AnyBuildOptions;
  readonly capacitor: IonicCapacitorOptions;
}
```

and also here:

```suggestion
    let serveOptions: AnyServeOptions | undefined;
    let buildOptions: AnyBuildOptions | undefined;
```

packages/@ionic/cli/src/commands/capacitor/run.ts
```diff
@@ -222,4 +229,39 @@ For Android and iOS, you can setup Remote Debugging on your device with browser
...
+        throw new FatalException(e.message);
+      }
+
+      throw e;
+    }
+  }
+}
+
+class CapacitorRunHook extends Hook {
+  readonly name = 'capacitor:run';
```
dwieeb, line 266: I wonder if we should `capacitor:run:after` and `capacitor:build:after` instead just so it's clear when they happen?

packages/@ionic/cli/src/commands/capacitor/run.ts
Heerschop: Created the union types, and updated the related code.

packages/@ionic/cli/src/commands/capacitor/run.ts
Heerschop: Changed the hook names as suggested :-)

packages/@ionic/cli/src/lib/hooks.ts
```diff
@@ -97,6 +104,25 @@ export abstract class Hook {
...
+    let environment: NodeJS.ProcessEnv = {};
+
+    for (const [key, value] of Object.entries(input)) {
+      if (typeof value === 'object') {
+        environment = {
+          ...environment,
+          ...this.generateCTXEnvironment(value, [...path, key]),
+        };
+      } else {
+        const name = [path, key].join('_');
```
dwieeb, line 118: Is this supposed to be this?

```suggestion
        const name = [...path, key].join('_');
```

dwieeb:
I don't know why I suggested `after`, I think `before` makes more sense. I'm sorry 😂 

The hook is technically being called _before_ building the app or running it on a device. We call it before we open the IDE, so it should probably be `before`. 😬 

Other than that, this is looking great!

Heerschop:
Ha ha, I agree and I have changed it to `before` :-)

Heerschop:
To make it complete I have added the `capacitor:run:after` and `capacitor:build:after` hook as well.

Heerschop:
Moved the **[INFO] Ready for use in your Native IDE!** message to after the invoke of the `capacitor:run:before` and `capacitor:build:before` hook.

The message log code is within the `options['open']` if statement, so now when you use the `--no-open` option, the message will not been shown anymore.

dwieeb:
I'm not sure if `capacitor:run:after` and `capacitor:build:after` should be included (yet). The Ionic CLI doesn't actually know when the app is run or built. If Capacitor ever gets a programmatic build, then the Ionic CLI could invoke it like it does for Cordova, in which case the hooks would make sense.

Thanks for fixing that other issue!

Heerschop:
It fine by me to remove or disable the `capacitor:run:after` and `capacitor:build:after` hooks. The use case it currently support is, doing pre and post processing when somebody is using the IDE (capacitor open). This might be a rare use case tho.

I have saved the work in a separate branch, and update this branch, so they are removed for now. If you changed your mind based on this comment, no problem let me know and I will restore them again :-)

dwieeb:
The only thing `capacitor open` does is tell the OS to open the IDE. We don't know when they run the app or when they're done using it.

dwieeb:
Awesome work @Heerschop! Thanks for the PR! I'll ship this ASAP.

Heerschop:
Thank you very much, it was nice working together.

PR #4451, opened 2020-05-22
feat(login): OpenID login flow is default login flow

giocalitri:


packages/@ionic/cli/src/lib/session.ts
```diff
@@ -133,6 +149,42 @@ export class ProSession extends BaseSession implements ISession {
...
+    const token = await oauthflow.exchangeRefreshToken(refreshToken);
+    await this.tokenLogin(token.access_token);
+    this.e.config.set('tokens.expiresInSeconds', token.expires_in);
+    this.e.config.set('tokens.issuedOn', (new Date()).toJSON());
+
+    return token.access_token;
+  }
 }
 
 export async function promptToLogin(env: IonicEnvironment): Promise<void> {
```
ltm, line 190: Should we replace `login()` with `webLogin()` in `promptToLogin()`?

packages/@ionic/cli/src/lib/oauth/openid.ts
```diff
@@ -0,0 +1,67 @@
...
+  OAuth2Flow,
+  OAuth2FlowDeps,
+  OAuth2FlowOptions,
+  TokenParameters
+} from './oauth';
+
+const AUTHORIZATION_URL = 'https://staging.ionicframework.com/oauth/authorize';
+const TOKEN_URL = 'https://api-staging.ionicjs.com/oauth/token';
+const CLIENT_ID = 'cli';
+const API_AUDIENCE = 'https://api.ionicjs.com';
```
ltm, line 17: These values should be configurable.

packages/@ionic/cli/src/commands/login.ts
```diff
@@ -157,16 +166,25 @@ If you are having issues logging in, please get in touch with our Support[^suppo
...
       this.env.log.nl();
 
       await this.env.session.ssoLogin(email);
     } else {
-      await this.env.session.login(email, password);
+      this.env.log.info(
+        `Ionic Login\n` +
+        `During this process, a browser window will open to authenticate you. Please leave this process running until authentication is complete.`
+      );
+      this.env.log.nl();
```
ltm, line 185: The user no longer gets to read this message before we send them to the browser. The old `ssoLogin` used to prompt for an email which would allow the user to see the message.

ltm: Maybe we can do a prompt similar to https://github.com/ionic-team/ionic-cli/blob/develop/packages/@ionic/cli/src/commands/link.ts#L382-L393

packages/@ionic/cli/src/lib/oauth/oauth.ts
```diff
@@ -22,53 +24,77 @@ export interface TokenParameters {
 }
 
 export interface OAuth2FlowOptions {
-  readonly authorizationUrl: string;
-  readonly tokenUrl: string;
-  readonly clientId: string;
+  readonly authorizationUrl?: string;
+  readonly tokenUrl?: string;
+  readonly clientId?: string;
```
ltm, line 29: I believe these three options are no longer used.

dwieeb:
Message when they try to run it non-interactively:

image

Message for people who use `--sso`:

image

packages/@ionic/cli/src/definitions.ts
```diff
@@ -509,10 +537,16 @@ export interface APIResponsePageTokenMeta extends APIResponseMeta {
 
 export type HttpMethod = 'GET' | 'POST' | 'PATCH' | 'PUT' | 'DELETE' | 'PURGE' | 'HEAD' | 'OPTIONS';
 
+export enum ContentType {
```
dwieeb, line 540:
```suggestion
export const enum ContentType {
```

`const enums` are preferred because they compile away

packages/@ionic/cli/src/lib/oauth/oauth.ts
```diff
@@ -22,53 +24,74 @@ export interface TokenParameters {
...
   }
 
+  async exchangeRefreshToken(refreshToken: string): Promise<T> {
+    const params = this.generateRefreshTokenParameters(refreshToken);
+    const { req } = await this.e.client.make('POST', this.oauthConfig.tokenUrl, this.accessTokenRequestContentType);
+
+    const res = await req.send(params);
+
+    // check the response status code first here
+    if (!res || !res.body || !res.status || res.status < 200 || res.status >= 300) {
```
dwieeb, line 77: I think you can use `!res.ok`: https://visionmedia.github.io/superagent/#response-status

packages/@ionic/cli/src/lib/oauth/oauth.ts
```diff
@@ -22,53 +24,74 @@ export interface TokenParameters {
...
+      throw new FatalException(
+        'API request was to refresh token was not successful.\n' +
```
dwieeb, line 79:
```suggestion
        'API request to refresh token was not successful.\n' +
```

packages/@ionic/cli/src/lib/session.ts
```diff
@@ -133,33 +147,61 @@ export class ProSession extends BaseSession implements ISession {
...
-  const password = await env.prompt({
-    type: 'password',
-    name: 'password',
-    message: 'Password:',
-    mask: '*',
-    validate: v => validators.required(v),
+  const login = await env.prompt({
+    type: 'confirm',
+    name: 'login',
+    message: 'Open the browser to login to your Ionic account?',
```
dwieeb, line 198:
```suggestion
    message: 'Open the browser to log in to your Ionic account?',
```

packages/@ionic/cli/src/lib/oauth/oauth.ts
```diff
@@ -22,53 +24,74 @@ export interface TokenParameters {
...
 
+  async exchangeRefreshToken(refreshToken: string): Promise<T> {
+    const params = this.generateRefreshTokenParameters(refreshToken);
+    const { req } = await this.e.client.make('POST', this.oauthConfig.tokenUrl, this.accessTokenRequestContentType);
+
+    const res = await req.send(params);
+
+    // check the response status code first here
+    if (!res || !res.body || !res.status || res.status < 200 || res.status >= 300) {
+      throw new FatalException(
```
dwieeb, line 78: How do users recover from this error?

giocalitri: probably the user needs to re-login to fix it: I'll add it to the message

PR #4459, opened 2020-06-04
feat: add pnpm support to build/serve commands

aparajita:


dwieeb:
This is more of a fix than a feat, wouldn't you say? This code was just missed when pnpm support was added.

aparajita:
@dwieeb Yeah, I guess it is really a fix. Whatever, as long as it's merged. 😁

PR #4511, opened 2020-08-12
feat(lab): improvements to the lab design

brandyscarney:
- Adds a device border around the previews, updates device ratio
- Updates the top statusbar on iOS and Android to use the notch (ios) and a new svg (android)
- Updates font from Avenir to Inter
- Improves UI in darker modes

Before & After

Screen Shot 2020-08-12 at 5 20 46 PM
Screen Shot 2020-08-12 at 5 31 16 PM


packages/@ionic/lab/stencil.config.js
```diff
@@ -8,6 +8,7 @@ exports.config = {
   ],
   copy: [
     { src: 'assets' },
+    { src: '../../../../../node_modules/@ionic-internal/ionic-ds/www/assets/fonts', dest: 'assets/fonts' },
```
imhoffd: You can use `require.resolve('@ionic-internal/ionic-ds')` to get the path and it will work no matter where it's installed

PR #4515, opened 2020-08-18
feat(): support vue projects

mhartington:


packages/@ionic/cli/src/lib/project/vue/serve.ts
```diff
@@ -14,34 +14,112 @@ export class VueServeRunner extends ServeRunner<VueServeOptions> {
...
+
+    if (strippedLine.includes('Compiled successfully')) {
+      this.emit('compile', this.chunks);
+      this.chunks = 0;
+    }
+
+    // const endBannerMsgs = ['development build', 'production build']
+    // if (endBannerMsgs.some(msg => strippedLine.includes(msg))) {
+    //   return false;
+    // }
```
liamdebeasi, line 85: do we still need this?

liamdebeasi (approved):

packages/@ionic/cli/src/lib/project/vue/serve.ts
mhartington: Mmmm...I haven't made up my mind yet. If we keep it, it will ignore some output, but it could be valuable information. I'd say for let's keep it for now then get feedback.

packages/@ionic/cli/src/commands/start.ts
```diff
@@ -539,9 +539,8 @@ Use the ${input('--type')} option to start projects using older versions of Ioni
     this.env.shell.alterPath = p => prependNodeModulesBinToPath(projectDir, p);
 
     if (!this.schema.cloned) {
-      if (this.schema.type === 'react') {
-        options['capacitor'] = true;
-      }
+      // Default to capacitor always
+      options['capacitor'] = true;
```
imhoffd: We probably should not do this for Angular yet until the starters are updated to remove the Cordova plugins and related code.

We will also need to improve the Cordova integration to install those plugins and advise the dev how to use the splash screen.

packages/@ionic/cli/src/lib/project/vue/serve.ts
```diff
@@ -14,34 +14,112 @@ export class VueServeRunner extends ServeRunner<VueServeOptions> {
...
+    // // warning, so don't set `HOST` if the host is set to `localhost`.
+    if (options.host !== DEFAULT_ADDRESS) {
+      env.HOST = options.host;
+    }
+
+    env.PORT = String(options.port);
+    env.HTTPS = options.https ? 'true' : 'false';
+
+    return { ...await super.buildEnvVars(options), ...env };
+  }
```
imhoffd, line 124: Pretty sure these will need to be updated to translate serve details to the underlying Vue Serve CLI. 

see https://cli.vuejs.org/guide/cli-service.html#vue-cli-service-serve

packages/@ionic/cli/src/commands/start.ts
mhartington: Agree. Will revert this and add a check for vue instead.

packages/@ionic/cli/src/lib/project/vue/serve.ts
mhartington: Good catch. Thanks.

PR #4591, opened 2020-10-15
chore(cd): wait 20 seconds to ensure npm command returns the correct tag after publish

nphyatt:


.github/workflows/cd.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           GIT_AUTHOR_EMAIL: hi@ionicframework.com
           GIT_COMMITTER_NAME: Ionitron
           GIT_COMMITTER_EMAIL: hi@ionicframework.com
+      - name: Sleep while NPM takes its time
```
imhoffd:
```suggestion
      - name: Sleep while npm takes its time
```

Issue #4392, opened 2020-04-15
Ionic Serve is stuck on "Files successfully emitted, waiting for typecheck results..."


Remling97:
**Description:**
When i execute "ionic serve" in my project, the output looks as follows:

> react-scripts.cmd start
[react-scripts] i ｢wds｣: Project is running at http://192.168.1.114/
[react-scripts] i ｢wds｣: webpack output is served from
[react-scripts] i ｢wds｣: Content not from webpack is served from C:\Users\Remling\Desktop\Code\hello\public
[react-scripts] i ｢wds｣: 404s will fallback to /
[react-scripts] Starting the development server...
[react-scripts]
[react-scripts] Files successfully emitted, waiting for typecheck results...

But thats where it stops, no site is opening, no error message, nothing.

**My `ionic info`:**

```
Ionic:

   Ionic CLI       : 6.5.0 (C:\Users\Remling\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/react 5.0.7

Capacitor:

   Capacitor CLI   : 2.0.1
   @capacitor/core : 2.0.1

Utility:

   cordova-res : not installed
   native-run  : 1.0.0

System:

   NodeJS : v13.13.0 (C:\Program Files\nodejs\node.exe)
   npm    : 6.14.4
   OS     : Windows 10
```

RichardPragnell:
I had the same problem (perhaps something already running on port 8000?) try using a different port:
ionic serve --port=8080

Remling97:
Hey, 
Thanks for the tip. tried it out, but it didnt work. 
I also tried a few different ports with the same results. 
Then i checked if something is already running on port 8000, but nothing is. 

If i start ionic serve, and check for the port again, the node process shows up as listening. So that should be fine, i dont understand why its not working

Remling97:
ok, i got some new insights to the error.
everything is working as it should. I can just open the page with localhost:8100 and it works just find, even the live reload. ionic serve just doesnt open the browser window by itself for some reason

WaelRabah:
Hey it seems to be a problem on Windows , same app on linux works fine but on windows I get this problem

stolenng:
I'm having the same on mac,
First compilation works good, after i change line of code it either stuck on:
`[react-scripts] Compiling...` or `[react-scripts] Files successfully emitted, waiting for typecheck results...`

Then my css is not reflected on browser and can't work like this at all :/

[...]

PR #4609, opened 2020-11-16
fix(cli): fix stdin to unblock react-scripts>3.4.0

ppoffice:
Should solve: https://github.com/ionic-team/ionic-cli/issues/4392

In `react-scripts 3.4.1`, a fragment of code for handling script exit was added:
https://github.com/facebook/create-react-app/blob/v3.4.1/packages/react-scripts/scripts/start.js#L168-L175

```js
if (isInteractive || process.env.CI !== 'true') {
  // Gracefully exit when stdin ends
  process.stdin.on('end', function() {
    devServer.close();
    process.exit();
  });
  process.stdin.resume();
}
```

Here, `react-scripts` tries to listen to the `stdin`, and it has no issues when you start `react-scripts` directly in Git Bash or CMD on Windows.
However, with `ionic serve`, the `@ionic/cli` will spawn `react-scripts` in a child process with its `stdin` exposed as a `pipe` (`{ stdio: 'pipe', ...`), which is not connected to any other input stream [like `stdout` and `stderr` do](https://github.com/ppoffice/ionic-cli/blob/e150aebf2628f52d42096266b71e95dede5ef872/packages/%40ionic/cli/src/lib/serve.ts#L537-L538).
And this causes the problem on Windows.
I'm not sure about the root cause, but changing `stdio: 'pipe'` to `stdio: ['inherit', 'pipe', 'pipe']` (let `react-scritps` inherit the `stdio` on `@ionic/cli`) seems to resolve the issue for me.

Please try this fix yourself and see if your issue is resolved.

Steps to reproduce the issue mentioned in https://github.com/ionic-team/ionic-cli/issues/4392:

1. Install Ionic CLI on Windows. I'm using `@ionic/cli@6.10.1`;
2. Create a new Ionic project with `ionic start myApp tabs --type=react`;
3. Upgrade `react-scripts` to `3.4.1` or above in your new Ionic project;
4. Run `ionic serve` from your project root and see if it hangs at `Files successfully emitted, waiting for typecheck results...`

Steps to fix the issue using this patch:

1. Go to wherever you installed `@ionic/cli` and open `lib\serve.js`;
2. Apply the fix in this PR;
3. Restart `ionic serve` to see if the issue is resolved.

### DIFF
--- a/packages/@ionic/cli/src/lib/serve.ts
+++ b/packages/@ionic/cli/src/lib/serve.ts
@@ -493,7 +493,7 @@ export abstract class ServeCLI<T extends ServeCLIOptions> extends EventEmitter {
   protected async spawn(options: T): Promise<void> {
     const args = await this.buildArgs(options);
     const env = await this.buildEnvVars(options);
-    const p = await this.e.shell.spawn(this.resolvedProgram, args, { stdio: 'pipe', cwd: this.e.project.directory, env: createProcessEnv(env) });
+    const p = await this.e.shell.spawn(this.resolvedProgram, args, { stdio: ['inherit', 'pipe', 'pipe'], cwd: this.e.project.directory, env: createProcessEnv(env) });
 
     return new Promise<void>((resolve, reject) => {
       const errorHandler = (err: NodeJS.ErrnoException) => {


imhoffd:
Nice work!

For context, this is the fix on react-scripts side: https://github.com/facebook/create-react-app/pull/8700

But it appears it was merged in for 4.0.0, not 3.x.

PR #4610, opened 2020-11-17
feat(capacitor): Capacitor 3 support

imhoffd:
TODO:
- [x] investigate editing android module's AndroidManifest.xml instead of app's
- [x] `ionic start` should use Capacitor 3

This PR adds support for Capacitor 3 in the Ionic CLI, necessary because of changes in Capacitor such as dynamic configuration and the availability of the `capacitor run` command.

This will require Capacitor version **3.0.0-alpha.7** or greater to test.

**Summary of Changes**
- `ionic cap run` will use `capacitor run` if available.
    - New `--target` option for targeting a specific device.
    - New `--list` option for listing target devices.
    - New `--open` option to use old flow of running `capacitor sync` and `capacitor open` (instead of `capacitor run`)
- `ionic cap run` will now edit the native platform's generated `capacitor.config.json` files to insert `server.url`.
- `ionic cap run` will no longer use Capacitor's `cleartext` option, but instead manually edit the app's `AndroidManifest.xml` and restore the file after the command is done.
    - `cleartext` is a Capacitor CLI option that takes effect during `capacitor sync`, but with dynamic configuration files we can no longer reliably edit the main config file.
- Capacitor configuration is first loaded via (hidden) `capacitor config` command, then falls back to trying to read `capacitor.config.json`.

### DIFF
--- a/packages/@ionic/cli/src/commands/capacitor/add.ts
+++ b/packages/@ionic/cli/src/commands/capacitor/add.ts
@@ -1,7 +1,10 @@
 import { validators } from '@ionic/cli-framework';
+import * as semver from 'semver';
 
 import { CommandInstanceInfo, CommandLineInputs, CommandLineOptions, CommandMetadata, CommandPreRun } from '../../definitions';
 import { input } from '../../lib/color';
+import { FatalException } from '../../lib/errors';
+import { pkgManagerArgs } from '../../lib/utils/npm';
 
 import { CapacitorCommand } from './base';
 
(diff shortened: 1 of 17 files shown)


dennisameling:
Just wanted to chime in here to mention that Ionic and this PR currently don't support the newly supported [Capacitor TypeScript config file](https://capacitorjs.com/docs/v3/config) (`capacitor.config.ts`). Details: https://github.com/ionic-team/ionic-cli/issues/4653

PR #4621, opened 2020-12-08
CT-434 handle git branches with '/' in name

lm3m:
* replacing only '/' and any of the valid git valid tokens in branch names that would cause issues

### DIFF
--- a/packages/@ionic/cli/src/commands/package/build.ts
+++ b/packages/@ionic/cli/src/commands/package/build.ts
@@ -387,7 +387,7 @@ This can be used only together with build type ${input('release')} for Android a
     if (!filename) {
       req.on('response', res => {
         const contentDisposition = res.header['content-disposition'];
-        filename = contentDisposition ? contentDisposition.split('=')[1] : 'output.bin';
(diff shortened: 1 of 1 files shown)


packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -387,7 +387,7 @@ This can be used only together with build type ${input('release')} for Android a
     if (!filename) {
       req.on('response', res => {
         const contentDisposition = res.header['content-disposition'];
-        filename = contentDisposition ? contentDisposition.split('=')[1] : 'output.bin';
+        filename = contentDisposition ? contentDisposition.split('=')[1].replace(/([/,?,<,>,*,|,\"])/g, '_') : 'output.bin';
```
ltm: Regular expression character sets don't use any separators:

```suggestion
        filename = contentDisposition ? contentDisposition.split('=')[1].replace(/([/?<>*|\"])/g, '_') : 'output.bin';
```

lm3m: ok. I defer to you on regex.

PR #4634, opened 2021-01-27
fix(Appflow): Fix issue with deploy add and deploy configure commands where sometimes neccessary variables weren't correctly set.

### LINKED ISSUE #4633: Ionic deploy add command doesn't perform required configuration
**Description:**

I've recently been fighting with my Ionic5/Angular/Capacitor app to try and get AppFlow working, but the tooling/documentation is lacking and it's been very difficult.

I've followed the quickstart guide: https://ionic.io/docs/appflow/quickstart/installation

The problem is, the 'ionic deploy add' command doesn't perform all the required configuration to make appflow work. This leads to a very frustrating experience where it's been very difficult to figure out what is missing to make things work and having to piece together things from forum posts like this one: https://forum.ionicframework.com/t/appflow-live-update-cant-make-it-work/185996/11

For Android, I had to add the following to 'strings.xml'

```
  <string name="ionic_max_versions">2</string>
  <string name="ionic_update_api">https://api.ionicjs.com</string>
  <string name="ionic_min_background_duration">30</string>
```

For iOS I've needed to amend info.plist:

```
<key>IonApi</key>
<string>$UPDATE_API</string>
<key>IonMaxVersions</key>
<string>$MAX_STORE</string>
<key>IonMinBackgroundDuration</key>
<string>$MIN_BACKGROUND_DURATION</string>
```

To hard coded values:

```
<key>IonApi</key>
<string>https://api.ionicjs.com</string>
<key>IonMaxVersions</key>
<string>2</string>
<key>IonMinBackgroundDuration</key>
<string>30</string>
```

I think the CLI command needs to be updated so that it checks these values are set, or that the $ variables that are meant to be substituted in here are set some where. I believe with cordova these would maybe be in config.xml and/or package.json but with Capacitor I don't think this is the case, or at least if it is, the 'ionic deploy add' command should take care of it.

I have to say, the troubleshooting link in the docs is next to useless, it just leads here: https://ionic.zendesk.com/hc/en-us

I would be a paying AppFlow subscriber months ago were it not for the fight to get this working. I've filled in the contact form a couple of times to see if someone from your sales team would get back to me about getting up n running on AppFlow, no one has. You've got two different prices listed on the site for the growth tier, $102 and $120, but without your sales team getting back to me, I can't work out which is correct to raise a P/O anyway.

Seriously, if you want to sell this, you need to support people who are trying to get it working, keep an eye on the forum and stack overflow and help people, get this working more smoothly. And contact people back who interested.

**Steps to Reproduce:**

**Output:**

Please see my forum post here for details of the error messages I get without performing the configuration above:

https://forum.ionicframework.com/t/live-updates-simply-not-appearing-on-device/203223

**My `ionic info`:**

```
Ionic:

   Ionic CLI                     : 6.12.3 (C:\Users\PaulMatthew\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 5.5.2
   @angular-devkit/build-angular : 0.1002.0
   @angular-devkit/schematics    : 10.2.0
   @angular/cli                  : 10.2.0
   @ionic/angular-toolkit        : 2.3.3

Capacitor:

   Capacitor CLI   : 2.4.6
   @capacitor/core : 2.4.6

Cordova:

   Cordova CLI       : 10.0.0
   Cordova Platforms : none
   Cordova Plugins   : no whitelisted plugins (0 plugins total)

Utility:

   cordova-res : not installed
   native-run  : not installed

System:

   Android SDK Tools : 26.1.1 (C:\Users\PaulMatthew\AppData\Local\Android\Sdk)
   NodeJS            : v14.15.4 (C:\Program Files\nodejs\node.exe)
   npm               : 6.14.10
   OS                : Windows 10
```

**Other Information:**
https://forum.ionicframework.com/t/live-updates-simply-not-appearing-on-device/203223

nphyatt:
fixes #4633

### DIFF
--- a/packages/@ionic/cli/src/commands/deploy/add.ts
+++ b/packages/@ionic/cli/src/commands/deploy/add.ts
@@ -2,7 +2,6 @@ import { MetadataGroup } from '@ionic/cli-framework';
 
 import { CommandInstanceInfo, CommandLineInputs, CommandLineOptions, CommandMetadata } from '../../definitions';
 import { input } from '../../lib/color';
-import { FatalException } from '../../lib/errors';
 import { runCommand } from '../../lib/executor';
 
 import { DeployConfCommand } from './core';
@@ -95,20 +94,12 @@ For Cordova projects it just takes care of running the proper Cordova CLI comman
   }
 
   async preRun(inputs: CommandLineInputs, options: CommandLineOptions): Promise<void> {
-    // check if it is already installed
-    const alreadyAdded = await this.checkDeployInstalled();
-    if (alreadyAdded) {
-      throw new FatalException(
-        `Appflow Deploy plugin is already installed.`
-      );
-    }
     // check if there are native integration installed
     await this.requireNativeIntegration();
     await this.preRunCheckInputs(options);
   }
 
-  async run(inputs: CommandLineInputs, options: CommandLineOptions, runinfo: CommandInstanceInfo): Promise<void> {
-    const integration = await this.getAppIntegration();
+  async addPlugin(options: CommandLineOptions, runinfo: CommandInstanceInfo, integration?: string) {
     if (integration === 'cordova') {
       let deployCommand = ['cordova', 'plugin', 'add', 'cordova-plugin-ionic'];
       const userOptions = this.buildCordovaDeployOptions(options);
@@ -125,6 +116,22 @@ For Cordova projects it just takes care of running the proper Cordova CLI comman
       );
       // install the plugin with npm
       await this.env.shell.run(installer, installerArgs, { stdio: 'inherit' });
+    }
+
+  }
+
+  async run(inputs: CommandLineInputs, options: CommandLineOptions, runinfo: CommandInstanceInfo): Promise<void> {
+    const integration = await this.getAppIntegration();
+
+    // check if it is already installed
+    const alreadyAdded = await this.checkDeployInstalled();
+    if (!alreadyAdded) {
+      this.addPlugin(options, runinfo, integration);
+    } else {
+      this.env.log.warn("Live Updates plugin already added. Reconfiguring only.");
+    }
+
+    if (integration === 'capacitor') {
       // generate the manifest
       await runCommand(runinfo, ['deploy', 'manifest']);
       // run capacitor sync
--- a/packages/@ionic/cli/src/commands/deploy/add.ts
+++ b/packages/@ionic/cli/src/commands/deploy/add.ts
@@ -126,7 +126,7 @@ For Cordova projects it just takes care of running the proper Cordova CLI comman
     // check if it is already installed
     const alreadyAdded = await this.checkDeployInstalled();
     if (!alreadyAdded) {
-      this.addPlugin(options, runinfo, integration);
+      await this.addPlugin(options, runinfo, integration);
     } else {
       this.env.log.warn("Live Updates plugin already added. Reconfiguring only.");
     }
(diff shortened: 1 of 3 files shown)


packages/@ionic/cli/src/commands/deploy/core.ts
```diff
@@ -289,6 +329,23 @@ export abstract class DeployConfCommand extends DeployCoreCommand {
         element.text = options[optionKey] as string;
       }
     }
+
+    // make sure required keys are set
+    // TODO
```
nphyatt:
```suggestion
```

kill the TODO note since it is now done.

packages/@ionic/cli/src/commands/deploy/core.ts
```diff
@@ -132,7 +151,7 @@ export abstract class DeployConfCommand extends DeployCoreCommand {
       return false;
     }
     if (!plistPath) {
-      this.env.log.info(`No Capacitor iOS project found.`);
+      this.env.log.warn(`No ${chalk.bold('Capacitor iOS')} project found you will need to rerun ${chalk.yellow('ionic deploy configure')} if you add it later.`);
```
imhoffd: Try not to use chalk directly--it can lead to people using different colors and be a nonsensical rainbow for users.

Instead, use the semantically named colors from `color.ts`: https://github.com/ionic-team/ionic-cli/blob/develop/packages/@ionic/cli/src/lib/color.ts

Commands are supposed to use the "input" color, which ends up being cyan, not yellow. Use "strong" for bold.

Also, these log messages can be more than one line. Most CLI info/warn/error messages have a "title" (the first line) and "message" (subsequent lines).

packages/@ionic/cli/src/commands/deploy/core.ts
```diff
@@ -237,7 +277,7 @@ export abstract class DeployConfCommand extends DeployCoreCommand {
       return false;
     }
     if (!stringXmlPath) {
-      this.env.log.info(`No Capacitor Android project found.`);
+      this.env.log.warn(`No ${chalk.bold('Capacitor Android')} project found you will need to rerun ${chalk.yellow('ionic deploy configure')} if you add it later.`);
```
imhoffd: Here, too.

packages/@ionic/cli/src/commands/deploy/add.ts
```diff
@@ -125,6 +116,22 @@ For Cordova projects it just takes care of running the proper Cordova CLI comman
...
+
+  }
+
+  async run(inputs: CommandLineInputs, options: CommandLineOptions, runinfo: CommandInstanceInfo): Promise<void> {
+    const integration = await this.getAppIntegration();
+
+    // check if it is already installed
+    const alreadyAdded = await this.checkDeployInstalled();
+    if (!alreadyAdded) {
+      this.addPlugin(options, runinfo, integration);
```
giocalitri: is this missing an await?

nphyatt: it would appear so yes good catch!

packages/@ionic/cli/src/commands/deploy/core.ts
giocalitri: @dwieeb updated

packages/@ionic/cli/src/commands/deploy/core.ts
giocalitri: @dwieeb updated here too

PR #4639, opened 2021-02-08
feat(appflow): Support downloading AAB, APK and WWW_ZIP from cli

jlane9:


packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -22,6 +22,9 @@ const ANDROID_BUILD_TYPES = ['debug', 'release'];
 const IOS_BUILD_TYPES = ['development', 'ad-hoc', 'app-store', 'enterprise'];
 const APP_STORE_COMPATIBLE_TYPES = ['release', 'app-store', 'enterprise'];
 const BUILD_TYPES = ANDROID_BUILD_TYPES.concat(IOS_BUILD_TYPES);
+const ANDROID_ARTIFACT_TYPES = ['AAB', 'APK'];
```
jlane9: API endpoint only supports uppercase versions of these, should I just `.toUpperCase()` whatever user input. Should I allow case insensitiveness in api?

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -325,8 +346,8 @@ This can be used only together with build type ${input('release')} for Android a
     }
   }
 
-  async getDownloadUrl(appflowId: string, buildId: number, token: string): Promise<DownloadUrl> {
-    const { req } = await this.env.client.make('GET', `/apps/${appflowId}/packages/${buildId}/download`);
+  async getDownloadUrl(appflowId: string, buildId: number, artifactType: string, token: string): Promise<DownloadUrl> {
+    const { req } = await this.env.client.make('GET', `/apps/${appflowId}/packages/${buildId}/download?artifactType=${artifactType}`);
```
jlane9: I could've solved the 'default' artifact type by making it optional here and let the api handle it. The api did have comments in it to eventually make it required.

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -276,6 +286,17 @@ This can be used only together with build type ${input('release')} for Android a
...
   }
 
+  async getDefaultArtifactType(platform: string): Promise<string> {
+    switch (platform) {
+      case 'ios':
+        return 'IPA';
+      case 'android':
+        return 'APK';
+      default:
+        throw new Error(`No default artifact type for platform '${platform}'`);
```
jlane9, line 379: Wasn't sure to do something simpler like `platform === 'ios' ? 'IPA' : 'APK'`, but did this incase we begin supporting other platforms e.g. `MacOS`, `Windows`, `Linux`.

packages/@ionic/cli/src/commands/package/build.ts
ltm: Yeah, let's just `.toUpperCase()` the input.

packages/@ionic/cli/src/commands/package/build.ts
ltm: 👍🏻

packages/@ionic/cli/src/commands/package/build.ts
ltm: 👍🏻

ltm (approved):

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -267,13 +313,49 @@ This can be used only together with build type ${input('release')} for Android a
...
+
+    return String(value);
+  }
+
+  async getDefaultArtifactType(platform: string): Promise<Array<string>> {
+    switch (platform) {
+      case 'ios':
+        return ['IPA', 'DSYM'];
+      case 'android':
+        return ['APK', 'AAB'];
```
jlane9: After some more thought I think we should default to only `IPA` and `APK`, that way if they were using `--build-file-name` before everything will still work the same going forward.

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -87,8 +96,9 @@ This can be used only together with build type ${input('release')} for Android a
         'android debug --environment="My Custom Environment Name"',
         'android debug --native-config="My Custom Native Config Name"',
         'android debug --commit=2345cd3305a1cf94de34e93b73a932f25baac77c',
+        'android debug --artifact-type=aab',
```
giocalitri, line 99: I would add an example with `--aab-name` or `--apk-name`

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -144,11 +154,45 @@ This can be used only together with build type ${input('release')} for Android a
         },
         {
           name: 'build-file-name',
-          summary: 'The name for the downloaded build file',
+          summary: '(Deprecated) The name for the downloaded build file',
```
giocalitri: I think the correct way is to use `MetadataGroup.DEPRECATED` in the groups and not the the description

jlane9: oh yeah

packages/@ionic/cli/src/commands/deploy/build.ts
```diff
@@ -209,4 +226,36 @@ Apart from ${input('--commit')}, every option can be specified using the full na
...
+        const apiErrorMessage = (e.response.body.error && e.response.body.error.message) ? e.response.body.error.message : 'Api Error';
+        throw new FatalException(`Unable to get download URL for build ${buildId}: ` + apiErrorMessage);
+      } else {
+        throw e;
+      }
+    }
+  }
+
+  async downloadBuild(url: string, filename: string): Promise<string> {
+    const { req } = await createRequest('GET', url, this.env.config.getHTTPConfig());
```
giocalitri, line 251: why not ` this.env.client.make`?

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -267,13 +315,49 @@ This can be used only together with build type ${input('release')} for Android a
...
+      let customBuildFileName = '';
+
+      if (options['build-file-name']) {
+        this.env.log.warn(`The ${input('--build-file-name')} option has been deprecated. Please use ${input('--ipa-name')}, ${input('--apk-name')} or ${input('--{ artifact }-name')}.`);
+        customBuildFileName = await this.sanitizeString(options['build-file-name']);
+      } else {
+        customBuildFileName = await this.sanitizeString(options[`${artifactType.toLowerCase()}-name`]);
+      }
+
+      const filename = await this.downloadBuild(url.url, customBuildFileName);
```
giocalitri, line 354: not sure this is necessary, but what about downloading the files in parallel?

giocalitri:
Please use conventional commits for the commit and PR title.

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -74,6 +77,12 @@ Customizing the build:
 Deploying the build to an App Store:
 - The ${input('--destination')} option can be used to deliver the app created by the build to the configured App Store. \
 This can be used only together with build type ${input('release')} for Android and build types ${input('app-store')} or ${input('enterprise')} for iOS.
+
+Downloading build artifacts:
+- By default once the build is complete, all artifacts are downloaded for the selected platform. ${input('aab')} and ${input('apk')} for Android \
+${input('ipa')} and ${input('dsym')} for iOS.
+- The ${input('--artifact-type')} option can be used to limit artifact downloads to only of that type. For instance, with Android, you can specify ${input('aab')} \
```
giocalitri, line 84: > limit artifact downloads to only of that type

is this correct?

jlane9: Yes, if you specify artifact type it will only download that type

giocalitri (approved):

PR #4641, opened 2021-02-10
feat(appflow): support Capacitor only apps with `ionic deploy manifest`

devinshoemaker:
This change is mainly for people who want to build their apps in Appflow. The Appflow runners call the `ionic deploy manifest` command during the build process, and we are currently updating Appflow to support non-Ionic Capacitor apps. This command is necessary for our flow, but only relies on the Capacitor build directory, so being able to fallback onto a Capacitor config works for this scenario.

### DIFF
--- a/packages/@ionic/cli/src/commands/deploy/capacitor.ts
+++ b/packages/@ionic/cli/src/commands/deploy/capacitor.ts
@@ -1,0 +1,66 @@
+import { BaseConfig } from "@ionic/cli-framework";
+import lodash = require("lodash");
+
+export interface CapacitorConfig {
+  appId?: string;
+  appName?: string;
+  webDir?: string;
+  server?: {
+    url?: string;
+    originalUrl?: string;
+  };
+}
+
+export interface CapacitorCLIConfig {
+  android: {
+    platformDirAbs: string;
+    srcMainDirAbs: string;
+    assetsDirAbs: string;
+  };
+  ios: {
+    platformDirAbs: string;
+    nativeTargetDirAbs: string;
+  };
+  app: {
+    extConfig: CapacitorConfig;
+  };
+}
+
+export class CapacitorJSONConfig extends BaseConfig<CapacitorConfig> {
+  constructor(p: string) {
+    super(p, { spaces: "\t" });
+  }
+
+  provideDefaults(config: CapacitorConfig): CapacitorConfig {
+    return config;
+  }
+
+  setServerUrl(url: string): void {
+    const serverConfig = this.get("server") || {};
+
+    if (typeof serverConfig.url === "string") {
+      serverConfig.originalUrl = serverConfig.url;
+    }
+
+    serverConfig.url = url;
+
+    this.set("server", serverConfig);
+  }
+
+  resetServerUrl(): void {
+    const serverConfig = this.get("server") || {};
+
+    delete serverConfig.url;
+
+    if (serverConfig.originalUrl) {
+      serverConfig.url = serverConfig.originalUrl;
+      delete serverConfig.originalUrl;
+    }
+
+    if (lodash.isEmpty(serverConfig)) {
+      this.unset("server");
+    } else {
+      this.set("server", serverConfig);
+    }
+  }
+}
--- a/packages/@ionic/cli/src/commands/deploy/manifest.ts
+++ b/packages/@ionic/cli/src/commands/deploy/manifest.ts
@@ -1,16 +1,25 @@
 import { MetadataGroup } from '@ionic/cli-framework';
+import { LOGGER_LEVELS } from '@ionic/cli-framework-output';
 import { map } from '@ionic/utils-array';
-import { readdirp, stat, writeFile } from '@ionic/utils-fs';
+import { pathExists, readdirp, stat, writeFile } from '@ionic/utils-fs';
 import { prettyPath } from '@ionic/utils-terminal';
 import * as crypto from 'crypto';
 import * as fs from 'fs';
+import lodash = require('lodash');
 import * as path from 'path';
+import * as Debug from 'debug';
 
 import { CommandMetadata } from '../../definitions';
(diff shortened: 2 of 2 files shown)


packages/@ionic/cli/src/commands/deploy/manifest.ts
```diff
@@ -22,19 +31,27 @@ export class DeployManifestCommand extends DeployCoreCommand {
   async getMetadata(): Promise<CommandMetadata> {
     return {
       name: 'manifest',
-      type: 'project',
+      type: 'global',
```
devinshoemaker, line 35: I had to switch this command to a type of `global` to avoid throwing an exception if `ionic.config.json` is not present.

packages/@ionic/cli/src/commands/deploy/capacitor.ts
```diff
@@ -0,0 +1,66 @@
+import { BaseConfig } from "@ionic/cli-framework";
```
devinshoemaker, line 1: To stay ahead of Capacitor 3 support coming soon, I copied some interface and class changes that will be coming later.

packages/@ionic/cli/src/commands/deploy/manifest.ts
```diff
@@ -22,19 +31,27 @@ export class DeployManifestCommand extends DeployCoreCommand {
...
     }
-    await this.requireNativeIntegration();
 
-    const buildDir = await this.project.getDistDir();
+    let distDir: string;
+    if (this.project) {
+      await this.requireNativeIntegration();
+      distDir = await this.project.getDistDir();
+    } else {
+      distDir = capacitorConfig!.webDir ? capacitorConfig!.webDir : 'www';
```
devinshoemaker: Even though we attempt to use the `cap config --json` command which should always return a `webDir` property, there's a chance that a user has not updated to Capacitor 3 so we still need to provide a fallback.

packages/@ionic/cli/src/commands/deploy/manifest.ts
giocalitri: I would put this as a comment in the code so it is clear why it is like this

packages/@ionic/cli/src/commands/deploy/manifest.ts
```diff
@@ -83,4 +100,53 @@ export class DeployManifestCommand extends DeployCoreCommand {
...
+    return path.resolve(this.env.ctx.execPath, CAPACITOR_CONFIG_JSON_FILE);
+  }
+
+  private getCapacitorCLIConfig = lodash.memoize(async (): Promise<CapacitorCLIConfig | undefined> => {
+    const args = ['config', '--json'];
+    const log = new Logger({
+      level: LOGGER_LEVELS.INFO,
+      handlers: createDefaultLoggerHandlers(),
+    });
+    const shell = new Shell({ log }, { alterPath: p => { return prependNodeModulesBinToPath(this.env.ctx.execPath, p)} });
```
devinshoemaker, line 116: I had to create a new shell to force `prependNodeModulesBinToPath`. If `ionic.config.json` is not present, then `this.env.shell` will not implement this, and the Capacitor command will fail.

packages/@ionic/cli/src/commands/deploy/capacitor.ts
giocalitri, line 1: Is this file temporary until the support of Cap 3 is merged?

devinshoemaker: Yes, exactly.

packages/@ionic/cli/src/commands/deploy/capacitor.ts
```diff
@@ -0,0 +1,66 @@
...
+import lodash = require("lodash");
```
giocalitri, line 2: is `lodash` already a dependency?

devinshoemaker: Yes, that's what's being used in the Capacitor 3 PR.

packages/@ionic/cli/src/commands/deploy/manifest.ts
```diff
@@ -22,19 +31,27 @@ export class DeployManifestCommand extends DeployCoreCommand {
...
-    const buildDir = await this.project.getDistDir();
+    let distDir: string;
+    if (this.project) {
+      await this.requireNativeIntegration();
+      distDir = await this.project.getDistDir();
+    } else {
+      distDir = capacitorConfig!.webDir ? capacitorConfig!.webDir : 'www';
+    }
+
+    const buildDir = distDir;
```
giocalitri: why are you assigning `distDir` to a constant?

devinshoemaker: I actually meant to clean that up, thanks for pointing that out.

devinshoemaker: This has been fixed.

packages/@ionic/cli/src/commands/deploy/manifest.ts
giocalitri: still, put this as comment in the code

packages/@ionic/cli/src/commands/deploy/capacitor.ts
giocalitri: ok

packages/@ionic/cli/src/commands/deploy/manifest.ts
devinshoemaker: Good idea, done.

packages/@ionic/cli/src/commands/deploy/manifest.ts
devinshoemaker: Done.

giocalitri (approved):

PR #4645, opened 2021-02-17
fix(appflow): Default should be all available not hard-coded by platform

jlane9:
Main issue here is that by default an `AAB` or `DSYM` artifact may not be available and under those circumstances we don't want the CLI to error because it tried to download every artifact that the platform _could_ produce. Instead download the artifacts that we _know_ the build produced by default.

Changes:
- `Artifact Type(s):` log output by default should not try to anticipate what artifacts we _could_ build and instead say `all available` or `not set` or something to let the user know we will download what we can.
- Default behavior uses the build output to find available artifact types instead of hard-coded.
- Multiple artifact requests delay failing only once all downloads are attempted
- Single artifact requests should work as it did before

### DIFF
--- a/packages/@ionic/cli/src/commands/package/build.ts
+++ b/packages/@ionic/cli/src/commands/package/build.ts
@@ -27,6 +27,12 @@ const IOS_ARTIFACT_TYPES = ['ipa', 'dsym'];
 const ARTIFACT_TYPES = ANDROID_ARTIFACT_TYPES.concat(IOS_ARTIFACT_TYPES);
 const TARGET_PLATFORM = ['Android', 'iOS - Xcode 11 (Preferred)', 'iOS - Xcode 10'];
 
+interface Artifact {
+  name: string;
+  url: string;
+  artifact_type: string;
+}
+
 export interface PackageBuild {
   job_id: number;
   id: string;
@@ -48,6 +54,7 @@ export interface PackageBuild {
   distribution_credential_name: string;
   job: any;
   distribution_trace: string;
+  artifacts: Artifact[];
 }
 
 interface DownloadUrl {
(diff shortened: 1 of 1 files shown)


packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -318,7 +315,7 @@ if you do not wish to download ${input('apk')}.
       ['Commit', strong(`${build.commit.sha.substring(0, 6)} ${build.commit.note}`)],
       ['Target Platform', strong(build.stack.friendly_name)],
       ['Build Type', strong(build.build_type)],
-      ['Artifact Type(s)', strong(`${artifactTypes}`)],
```
jlane9, line 321: I changed this and the above code, since `AAB` or `DSYM` may not always be available by default.

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -369,17 +370,6 @@ if you do not wish to download ${input('apk')}.
     return String(value);
   }
 
-  async getDefaultArtifactType(platform: string, useOriginalDefault?: boolean): Promise<Array<string>> {
```
jlane9, line 372: This was deleted to instead download all available artifacts that the build has available to it.

packages/@ionic/cli/src/commands/package/build.ts
```diff
@@ -335,24 +332,28 @@ if you do not wish to download ${input('apk')}.
...
 
-      let customBuildFileName = '';
-
-      if (options['build-file-name']) {
-        this.env.log.warn(`The ${input('--build-file-name')} option has been deprecated. Please use ${input('--ipa-name')}, ${input('--apk-name')} or ${input('--{ artifact }-name')}.`);
-        customBuildFileName = await this.sanitizeString(options['build-file-name']);
-      } else {
-        customBuildFileName = await this.sanitizeString(options[`${artifactType.toLowerCase()}-name`]);
+      if (errors.length) {
+        errors.forEach((error) => this.env.log.error(error.message));
```
jlane9: Still need to figure out exactly how we want to handle the error.

jlane9: Probably need this to be a warning and throw an exception after that says "Some artifacts failed to download" or something similar so we can output each erroneous artifact.

devinshoemaker (approved):

giocalitri (approved):

Issue #3965, opened 2019-05-03
ionic serve --lab always tries to access localhost


t91710:
**Description:**
I run ionic lab on a dev server on my local network. When upgrading to latest ionic-cli version it's no longer possible to connect from ionic-lab to the angular dev server since it always tries to connect to localhost instead of the address of the server. The problem is that localhost is hard coded in
https://github.com/ionic-team/ionic-cli/blob/44b091800aa36e5b16206bf7c0e1a77caa8480e6/packages/ionic/src/lib/serve.ts#L715
It should probably look at the address of the command which ionic serve is started with, and if there is something else than 0.0.0.0 (which is also default), this address should be used when starting ionic-lab. 

**Steps to Reproduce:**
1. Get two computers
2. Get the ip of the first machine, for example 192.168.1.1
3. Run ionic serve --address 192.168.1.1 --lab --lab-host 0.0.0.0 -- --disableHostCheck
4. From your other computer, try to access 192.168.1.1:8200 (ionic lab) -> Does not work, because it tries to connect to localhost:8100 instead of 192.168.1.1:8100

**My `ionic info`:**
```
Ionic:

   ionic (Ionic CLI)             : 4.12.0 (/usr/lib/node_modules/ionic)
   Ionic Framework               : @ionic/angular 4.2.0
   @angular-devkit/build-angular : 0.13.8
   @angular-devkit/schematics    : 7.2.4
   @angular/cli                  : 7.3.8
   @ionic/angular-toolkit        : 1.4.1

Cordova:

   cordova (Cordova CLI) : not installed
   Cordova Platforms     : not available
   Cordova Plugins       : not available

System:

   NodeJS : v10.15.3 (/usr/bin/node)
   npm    : 6.4.1
   OS     : Linux 3.10
```

**Other Information:**
Should be easy to just use the correct variable from the ionic serve command on this row (https://github.com/ionic-team/ionic-cli/blob/44b091800aa36e5b16206bf7c0e1a77caa8480e6/packages/ionic/src/lib/serve.ts#L715) instead of hardcoded localhost-string. The ionic-lab bin already supports this, I have tested to do the change locally.

dwieeb:
I never thought someone would do what you're doing 😆 

No reason this can't be fixed this up, though!

t91710:
I understand it's not a very common thing to do. But I do it this way to please my project managers. They want to be able to look at the app progress, and I don't want to show it every time :smile:

[...]

PR #4650, opened 2021-02-24
fix(#3965): give --lab access to serve externalAddress

Ghustavh97:
This should fix a bug where lab is trying to always grab content
from localhost even if the content is hosted on a different address.

This should help with accessing lab on an external network in a browser
when say for extample the ionic-cli is running on ubuntu server, where a
browser is not available.

The reason we will use `serveDetails.externalAddress` instead of
`labDetails.host` is because `serve` is where we will be getting our content
from and by default is hosted on a seperate port from lab.

This change should not break anything since `serveDetails.externalAddress`
will default to the initial `localhost` if no --host argument is passed.

## Before

nw-console
nw-browser

## After

w-console
w-browser

### DIFF
--- a/packages/@ionic/cli/src/lib/serve.ts
+++ b/packages/@ionic/cli/src/lib/serve.ts
@@ -672,7 +672,7 @@ class IonicLabServeCLI extends ServeCLI<IonicLabServeCLIOptions> {
 
     const pkg = await this.e.project.requirePackageJson();
 
-    const url = `${serveDetails.protocol}://localhost:${serveDetails.port}`;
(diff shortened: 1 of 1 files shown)


ashamim92 (approved):

liamdebeasi:
Thanks for the PR and apologies for the delay. We appreciate the work you put into creating this PR. We recently deprecated Ionic Lab in favor of using the built in browser developer tools for mobile emulation: https://github.com/ionic-team/ionic-cli/pull/4930. I am going to close this PR as we are not adding new Ionic Lab functionality at this time. Please let me know if you have any questions.

PR #4690, opened 2021-05-14
Don't create capacitor.config.json if it doesn't exist

jcesarmobile:
`personalize` method was supposed to edit the `capacitor.config.json` config values, but since capacitor creates a .ts file now, the CLI is creating a new `capacitor.config.json` file, so this will check if the exists and overwrite on it only if it exists, but don't create a new one.

But I don't think we need the `personalize` method at all as the `appName` and `appId` are already passed to capacitor init command. I've tried to look into it and ended in commits that din't explain why it was included, so didn't remove it just in case it's used somewhere else.


mhartington:
Merged via https://github.com/ionic-team/ionic-cli/commit/9b88d3a9de545881364e87f99b00a520e257d7c0

PR #4706, opened 2021-06-01
fix: update package-lock.json when the --no-deps flag is used

jaredhagen:
updated `ionic start` to update the package-lock.json when the --no-deps flag is passed


giocalitri:
is this an actual `fix` meaning the previous behavior was a `bug`?

devinshoemaker (approved):

jaredhagen:
I'd consider the previous behavior a bug.  It resulted in an out-of-sync `package-lock.json`.  It's certainly an unintended result.

devinshoemaker (approved):

PR #4745, opened 2021-08-05
chore: Cap is no longer experimental and is default for Angular

nphyatt:


jcesarmobile (commented):
I think you can now delete the block where it used to prompt

https://github.com/ionic-team/ionic-cli/blob/8a84ec2081e6e6193cbd1c724f020b83a3d0df5c/packages/%40ionic/cli/src/commands/start.ts#L562-L573

ziftytodd:
This change now installs capacitor on all starter projects. Maybe it's just me, but I don't want capacitor on my web only projects. Please bring back the prompt.

PR #4748, opened 2021-08-15
fix: add missing 'no-'s to command metadata

EthanOrlander:
Some options are missing the "no-" part. I've added it in where it's missing.

Also wondering: where there are discrepancies between the docs and the help command, which is right? The docs don't contradict the help command, but they do include some options that are not present in the help command. I can go in and add these if helpful.


jcesarmobile:
Thanks for the PR, but when a boolean option is set to true as default, the CLI automatically adds the `--no-` before the name, so with your changes instead of `--no-build` it appears as `--no-no-build`.

PR #4758, opened 2021-09-02
fix: support Capacitor 1 apps

devinshoemaker:
The Ionic CLI attempts to get the users Capacitor config with the `npx cap config --json` command that was introduced in Capacitor 3. This code assumes that if the command does not work then an output (stdout) would not be returned. This is the case with Capacitor 2 as it returns the "command not found" error to stderror. However, Capacitor 1 returns this to stdout which then causes an error because there is an output, but it not JSON and not parsable.

### DIFF
--- a/packages/@ionic/cli/src/lib/integrations/capacitor/index.ts
+++ b/packages/@ionic/cli/src/lib/integrations/capacitor/index.ts
@@ -195,9 +195,16 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
     if (!output) {
       debug('Could not get config from Capacitor CLI (probably old version)');
       return;
+    } else {
+      try {
+        // Capacitor 1 returns the `command not found` error in stdout instead of stderror like in Capacitor 2
+        // This ensures that the output from the command is valid JSON to account for this
+        return JSON.parse(output);
+      } catch(e) {
+        debug('Could not get config from Capacitor CLI (probably old version)', e);
+        return;
+      }
     }
-
-    return JSON.parse(output);
   });
 
   getCapacitorConfig = lodash.memoize(async (): Promise<CapacitorConfig | undefined> => {


giocalitri (approved):

packages/@ionic/cli/src/lib/integrations/capacitor/index.ts
```diff
@@ -195,9 +195,16 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
     if (!output) {
       debug('Could not get config from Capacitor CLI (probably old version)');
       return;
+    } else {
+      try {
+        // Capacitor 1 returns the `command not found` error in stdout instead of stderror like in Capacitor 2
+        // This ensures that the output from the command is valid JSON to account for this
+        return JSON.parse(output);
+      } catch(e) {
+        debug('Could not get config from Capacitor CLI (probably old version)', e);
```
smzelek, line 204: This could potentially silence legitimate JSON errors which would've previously halted the execution.
We could check if the error is an instance of [SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions) if we want to still properly bail out when a syntax error is found in the JSON, and perhaps check the running version of Capacitor instead to purposefully swallow Cap1 errors, or check for a known substring in `output`, e.g. `output.contains("command not found")` ?

smzelek (changes requested):
Let me know what you think.

packages/@ionic/cli/src/lib/integrations/capacitor/index.ts
devinshoemaker: I think that makes sense, but would there be a difference in behavior? Right now if we can't get the Capacitor config from the CLI then we fallback to trying to read a `capacitor.config.json`, and I think that would still be the same behavior with the changes you proposed.

smzelek: Sounds good!

smzelek (approved):

PR #4772, opened 2021-10-06
Update capacitor install text and make it default true

mlynch:


mlynch:
Whoops, latest already defaults this to on 🚀

PR #4776, opened 2021-10-13
feat(appflow): deprecate appflow functionality in @ionic/cli

smzelek:


packages/@ionic/cli/src/lib/updates.ts
```diff
@@ -77,6 +77,13 @@ export async function runUpdateCheck({ config }: PersistPackageVersionsDeps): Pr
   await writeUpdateConfig(dir, newUpdateConfig);
 }
 
+export const IONIC_CLOUD_CLI_MIGRATION = () =>
```
jlane9: Does this really need to be a function?

smzelek: No, it just made the formatting easier to read for free is all.

jaredhagen: Maybe just turn it into an IIFE: https://developer.mozilla.org/en-US/docs/Glossary/IIFE

giocalitri (approved):

jaredhagen (approved):

PR #4820, opened 2022-03-02
(fix #4819) Allow ionic login to work in a non-interactive process

### LINKED ISSUE #4819: Allow ionic login to work in a non-interactive process
Allow ionic login to work in a non-interactive process. This could be done using the `--confirm` flag which should "_Turn off interactive prompts_"

dtarnawsky:
Allow ionic login to work in a non-interactive process. This could be done using the --confirm flag which should "Turn off interactive prompts"

### DIFF
--- a/packages/@ionic/cli/src/commands/login.ts
+++ b/packages/@ionic/cli/src/commands/login.ts
@@ -160,7 +160,7 @@ If you are having issues logging in, please get in touch with our Support[^suppo
       await this.logout();
       await this.env.session.login(email, password);
     } else {
-      if (!this.env.flags.interactive) {
+      if (!this.env.flags.interactive && !this.env.flags.confirm) {
         throw new FatalException(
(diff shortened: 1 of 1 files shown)


packages/@ionic/cli/src/commands/login.ts
```diff
@@ -170,12 +170,15 @@ If you are having issues logging in, please get in touch with our Support[^suppo
...
-      });
+      let login = true;
+      if (!this.env.flags.confirm) {
+        login = await this.env.prompt({
+          type: 'confirm',
+          name: 'continue',
+          message: 'Open the browser to log in to your Ionic account?',
+          default: true,
+        });
+      }
```
ltm: Is this change actually needed? I believe the prompt will use the default value in non-interactive mode.

dtarnawsky: Looks like it is not. I've removed the unnecessary check

ltm (approved):

PR #4823, opened 2022-03-09
feat(integrations): add ng add cordova-builder

mhartington:
This modifies the Cordova integration to run `ng add @ionic/cordova-builders` for angular projects.

For new Ionic projects, we have removed the old Cordova builders in favor of moving forward with Capacitor.

For legacy projects, this will add back the builders for Angular app and install the necessary dependencies.

### DIFF
--- a/packages/@ionic/cli/src/lib/integrations/cordova/index.ts
+++ b/packages/@ionic/cli/src/lib/integrations/cordova/index.ts
@@ -120,6 +120,14 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
       },
     });
     await super.add(details);
+    if(this.e.project.type === 'angular'){
+      try {
+      const integration = this.e.project.requireIntegration(this.name);
+      await this.e.shell.run('ng', ['add', '@ionic/cordova-builders', '--skip-confirmation'], {cwd: integration.root});
+      } catch (e) {
+        debug('Error while adding @ionic/cordova-builders.');
+      }
+    }
   }
 
   async getCordovaConfig(): Promise<configlib.CordovaConfig | undefined> {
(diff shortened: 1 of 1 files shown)


packages/@ionic/cli/src/lib/integrations/cordova/index.ts
```diff
@@ -120,6 +120,14 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
       },
     });
     await super.add(details);
+    if(this.e.project.type === 'angular'){
```
liamdebeasi:
```suggestion
    if (this.e.project.type === 'angular') {
```

packages/@ionic/cli/src/lib/integrations/cordova/index.ts
```diff
@@ -120,6 +120,14 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
...
+      try {
+      const integration = this.e.project.requireIntegration(this.name);
+      await this.e.shell.run('ng', ['add', '@ionic/cordova-builders', '--skip-confirmation'], {cwd: integration.root});
```
liamdebeasi:
```suggestion
        const integration = this.e.project.requireIntegration(this.name);
        await this.e.shell.run('ng', ['add', '@ionic/cordova-builders', '--skip-confirmation'], { cwd: integration.root });
```

liamdebeasi (changes requested):

packages/@ionic/cli/src/lib/integrations/cordova/index.ts
```diff
@@ -120,6 +120,14 @@ export class Integration extends BaseIntegration<ProjectIntegration> {
...
+      } catch (e) {
+        debug('Error while adding @ionic/cordova-builders.');
```
liamdebeasi: Can we pass through the error here? Otherwise the error message by itself is not really helpful

liamdebeasi (changes requested):

liamdebeasi (approved):

PR #4669, opened 2021-04-14
extra args passed to vue-cli

aperron:
A small fix to pass extra args to vue-cli-service when running the 'build' command

Example: `ionic build -- --mode staging` => `vue-cli-service build --mode staging`

I take example on method [buildOptionsToNgArgs](https://github.com/ionic-team/ionic-cli/blob/5ca557bad2e04a147fc5aed9fbf0298def628a0e/packages/%40ionic/cli/src/lib/project/angular/build.ts#L125) for angular projects.

resolve #4642

PR #4827, opened 2022-03-16
extra args passed to vue-cli serve

mwahlhuetter:
Applies the same fix for additional args for the serve command as #4669


jcesarmobile:
looks good, thanks!

PR #4832, opened 2022-03-31
Make sure input has correct format in generateCTXEnvironment

bboldi:
I ran into some cases where input was undefined, and ionic run failed with `TypeError: Cannot convert undefined or null to object` , I propose making sure that input will not break the build process if there's something wrong with json config that it uses values from to run hooks.


bboldi:
Did I miss something? Do I need to do anything else to get this reviewed? Thanks!

PR #4837, opened 2022-04-24
remove deprecated @types packages

beenotung:
open and os-name have built-in typescript support, no longer need to use the @types packages


jcesarmobile (approved):
looks good, thanks!

PR #4851, opened 2022-05-24
feat: automatically toggle iOS NSAppTransportSecurity settings during live reload

theproducer:


giralte-ionic (approved):

IT-MikeS (approved):

packages/@ionic/cli/src/commands/capacitor/base.ts
```diff
@@ -236,6 +249,17 @@ export abstract class CapacitorCommand extends Command {
 
         manifest.enableCleartextTraffic();
         await manifest.save();
+      }      
+
+      if (platform === 'ios') {
```
jcesarmobile:
```suggestion
      if (platform === 'ios' && !options["external"]) {
```
when passing `--external` option it would use a local IP to load the app and ATS won't block the connection, so would be good to not edit the plist if it's not necessary

jcesarmobile (approved):

IT-MikeS (approved):

PR #4854, opened 2022-05-26
feat(cli): allow cap electron commands

jcesarmobile:
Since electron is now a community platform, when `ionic cap add electron` command is run, ionic CLI will try to install @capacitor/electron 3.x, which no longer exists.

This PR will change the install behaviour to only install core platforms.
It will also add electron to the installed platforms array so users can use ionic cap electron command (if not on the array the command is still forwarded to capacitor CLI, but shows an ugly error, this prevents the error)

Reference https://forum.ionicframework.com/t/no-matching-version-found-for-capacitor-electron-3-5-1/223508

Alternatives:
The CLI could install @capacitor-community/electron@latest if `ionic cap add electron` command is run
The CLI could error if electron or other community platforms are used with `ionic cap` command with a message that indicates that those platforms are not supported by ionic CLI and they should use the capacitor commands without ionic "prefix", instead of forwarding them to capacitor CLI

### DIFF
--- a/packages/@ionic/cli/src/commands/capacitor/base.ts
+++ b/packages/@ionic/cli/src/commands/capacitor/base.ts
@@ -134,6 +134,11 @@ export abstract class CapacitorCommand extends Command {
     }
   });
 
+  isCorePlatform(platform: string): boolean {
+    const platforms = ['android', 'ios'];
+    return platforms.includes(platform);
+  }
+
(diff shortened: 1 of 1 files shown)


IT-MikeS (approved):
Just having the command forwarded to the Capacitor CLI should be fine, then users can just use `@capacitor-community/electron` in place of `electron` like we suggest in the Capacitor CLI

PR #4928, opened 2022-11-08
Inline Source Maps

### LINKED ISSUE #3802: Add ability to debug Ionic4 TypeScript files when running on device or emulator
As per this issue (https://github.com/ionic-team/ionic/issues/16455#issuecomment-449962914), would like to be able to debug when running on an emulator as could in previous versions.

As mentioned in the link, this will work with inline source maps but currently have to edit and Angular build source file to do this. 

Would like this out of the box

Cybertron01Z:
since @capacitor/cli 4.2.0 the --inline option is available to avoid .js.map files. This is necessary to debug typescript code on Android

closes #3802

### DIFF
--- a/packages/@ionic/cli/src/commands/capacitor/copy.ts
+++ b/packages/@ionic/cli/src/commands/capacitor/copy.ts
@@ -17,9 +17,9 @@ export class CopyCommand extends CapacitorCommand implements CommandPreRun {
       },
       {
         name: 'inline',
-        summary: 'Use inline source maps (only available on capacitor 4.2.0)',
+        summary: 'Use inline source maps (only available on capacitor 4.2.0+)',
         type: Boolean,
-        default: true
+        default: false
       }
     ];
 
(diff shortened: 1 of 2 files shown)


IT-MikeS (approved):

packages/@ionic/cli/src/commands/capacitor/copy.ts
```diff
@@ -14,6 +15,12 @@ export class CopyCommand extends CapacitorCommand implements CommandPreRun {
         type: Boolean,
         default: true,
       },
+      {
+        name: 'inline',
+        summary: 'Use inline source maps (only available on capacitor 4.2.0)',
+        type: Boolean,
+        default: true
```
dtarnawsky: The default should be false right @jcesarmobile @IT-MikeS ?
That way if you omit the option it won't inline source code and if you specify --inline it will work

jcesarmobile: yeah, should be false.
making it true makes the `--help` parameter to show it as `--no-inline` instead of `--inline` and the description makes no sense then.
I've updated it

jcesarmobile (approved):

dtarnawsky (approved):

PR #4972, opened 2023-03-22
chore!: bump required node version to 16

jcesarmobile:


packages/@ionic/utils-subprocess/src/index.ts
```diff
@@ -187,16 +187,16 @@ export class Subprocess {
 
         if (code === 0) {
           return resolve();
-        }
-
-        if (signal) {
+        } else  if (signal) {
```
liamdebeasi:
```suggestion
        } else if (signal) {
```

packages/@ionic/discover/src/publisher.ts
```diff
@@ -173,6 +173,8 @@ export function prepareInterfaces(interfaces: { [index: string]: os.NetworkInter
 
       return false;
     });
+  }
+
```
liamdebeasi:
```suggestion
```

liamdebeasi (approved):
Two whitespace fixes, but otherwise this looks good.

jcesarmobile:
I would have expected `npm run lint:fix` to fix those, but `npm run lint` just returns true, so it does nothing.
good catch!

PR #4974, opened 2023-03-23
chore(deps): update typescript to 4.8

jcesarmobile:
I wanted to update to typescript 5 but it has problems with jest
So I tried 4.9 and it has problems with our version of @types/node
So I've updated to 4.8


mhartington (approved):
Looks good to me

liamdebeasi (approved):

PR #4981, opened 2023-03-28
chore: go back to commonjs

jcesarmobile:


tsconfig.base.json
```diff
@@ -3,7 +3,7 @@
     "allowUnreachableCode": false,
     "declaration": true,
     "importHelpers": true,
-    "module": "node16",
+    "module": "CommonJS",
```
liamdebeasi, line 6:
```suggestion
    "module": "commonJS",
```

liamdebeasi (changes requested):
Should we just be reverting this commit as a whole? https://github.com/ionic-team/ionic-cli/commit/e86958ac98e2846b590cfdaf7b32dce99076323e#diff-0b280a445be167f54cd916c0718c952b8e0d4ca9621903d05eec25efd4c6ed6d

PR #4991, opened 2023-03-30
fix(cli): ionic info returns package information

sean-perkins:
## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Developers running `ionic info` against an `@ionic/angular` application on v7 will receive the following warning:

```
[WARN] Error loading @ionic/angular package.json: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package' is
       not defined by "exports" in /Users/ken/Projects/Training/tea-taster-ng/node_modules/@ionic/angular/package.json

```

Issue URL: https://github.com/ionic-team/ionic-cli/issues/4992

## What is the new behavior?

- Updates the utility for reading the `package.json` to include the full path
- `ionic info` returns the correct information for a project in both v6 and v7 (tested locally)

### DIFF
--- a/packages/@ionic/cli/src/lib/project/index.ts
+++ b/packages/@ionic/cli/src/lib/project/index.ts
@@ -508,7 +508,7 @@ export abstract class Project implements IProject {
     let pkgPath: string | undefined;
 
     try {
-      pkgPath = pkgName ? require.resolve(`${pkgName}/package`, { paths: compileNodeModulesPaths(this.directory) }) : this.packageJsonPath;
+      pkgPath = pkgName ? require.resolve(`${pkgName}/package.json`, { paths: compileNodeModulesPaths(this.directory) }) : this.packageJsonPath;
       pkg = await readPackageJsonFile(pkgPath);
     } catch (e: any) {
       if (logErrors) {
--- a/packages/@ionic/cli/src/lib/project/index.ts
+++ b/packages/@ionic/cli/src/lib/project/index.ts
@@ -521,7 +521,7 @@ export abstract class Project implements IProject {
 
   async requirePackageJson(pkgName?: string): Promise<PackageJson> {
     try {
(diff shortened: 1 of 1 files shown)


packages/@ionic/cli/src/lib/project/index.ts
```diff
@@ -508,7 +508,7 @@ export abstract class Project implements IProject {
     let pkgPath: string | undefined;
 
     try {
-      pkgPath = pkgName ? require.resolve(`${pkgName}/package`, { paths: compileNodeModulesPaths(this.directory) }) : this.packageJsonPath;
+      pkgPath = pkgName ? require.resolve(`${pkgName}/package.json`, { paths: compileNodeModulesPaths(this.directory) }) : this.packageJsonPath;
```
liamdebeasi, line 511: Will this break for developers on older versions of Ionic?

sean-perkins: I tested on Ionic 6 and 7 for Angular, React and Vue and it continued to log the appropriate message (no warning displayed). I don't have an Ionic 5 project to test against, but think we are good as long as the CLI supports the current supported versions of Ionic Framework. It seems to support the changes in v7 with the `exports` object in the package.json as well as the legacy format from earlier versions of ng-packagr.

liamdebeasi: Do we need to make the same change in `requirePackageJson` below?

sean-perkins: I guess it wouldn't hurt - I'm not sure what the part of the implementation accounts for. It looks to be potentially for Angular workspace configurations (having many apps in one project).

liamdebeasi (approved):

PR #5001, opened 2023-04-14
chore(): clean up lingering ionic v1 and v3 refs

liamdebeasi:
We removed support for legacy v1 and v3 projects in Ionic CLI v7. However, there are several lingering references to these projects in the CLI. This PR removes those lingering references.


mapsandapps (approved):
Looks good!

PR #5005, opened 2023-04-28
feat(): add standalone as an option

liamdebeasi:
This PR introduces an option to the Ionic CLI to create an Angular app with either NgModules or Standalone. Developers will be prompted to choose one and will receive the appropriate starter based on their selection. 

Related: https://github.com/ionic-team/starters/pull/1773

The NgModules option will pull the `angular` starter app (named this way for backwards compatibility with older versions of the CLI). The Standalone option will pull the `angular-standalone` starter app.

I split out the standalone support into its own project type since that is what we did for the Vite starters.

### DIFF
--- a/packages/@ionic/cli/src/lib/start.ts
+++ b/packages/@ionic/cli/src/lib/start.ts
@@ -2,7 +2,7 @@ import { readJson } from '@ionic/utils-fs';
 import { columnar } from '@ionic/utils-terminal';
 import * as lodash from 'lodash';
 
-import { COLUMNAR_OPTIONS, PROJECT_TYPES } from '../constants';
+import { COLUMNAR_OPTIONS, PROJECT_TYPES, ANGULAR_STANDALONE } from '../constants';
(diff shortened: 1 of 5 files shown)


packages/@ionic/cli/src/lib/start.ts
```diff
@@ -272,6 +272,41 @@ export const STARTER_TEMPLATES: StarterTemplate[] = [
...
+    description: 'A starting project with a list',
+    id: 'angular-standalone-official-list',
+  },
+  {
+    name: 'my-first-app',
+    projectType: ANGULAR_STANDALONE,
+    type: 'repo',
+    description: 'A template for the "Build Your First App" tutorial',
+    repo: 'https://github.com/ionic-team/photo-gallery-capacitor-ng',
+  },
```
liamdebeasi, line 309: Note: This project does not ship with standalone components, but I added it so we don't get an error on build

sean-perkins (approved):

PR #5036, opened 2023-08-29
Removed device selection in "open" command

nicorac:
There's no need to show a device selection in `open` command when there's only one device available.

Since there's only a single possible selection, blocking `open` command waiting for user answer is worthless 😉.

NOTE: the behavior added by this PR to **ionic cli** is already present in **capacitor cli**.


nicorac:
Just rebased my patch, but workflow won't restart...🙄

PR #5063, opened 2023-11-16
Update definitions.ts to support bun.sh

ekoeryanto:


slogivewd:
Its not that easy, we need to add support for bun through this file too https://github.com/ionic-team/ionic-cli/blob/develop/packages/%40ionic/cli/src/lib/utils/npm.ts

slogive:
I've created this PR to support it, works as expected https://github.com/ionic-team/ionic-cli/pull/5097

PR #5068, opened 2023-12-12
chore: deprecate the ssl command

liamdebeasi:
This SSL command is largely redundant now that framework tooling has better support for SSL.

Developers using Vite should leverage the [http config option](https://vitejs.dev/config/server-options.html#server-https) as well as [@vitejs/plugin-basic-ssl](https://github.com/vitejs/vite-plugin-basic-ssl).

Developers using Angular CLI should leverage the [`--ssl` flag](https://angular.io/cli/serve) when serving. Alternatively, developers can update `angular.json` to provide this configuration.

Developers can also use https://github.com/FiloSottile/mkcert to generate locally-trusted development certificates. See https://web.dev/articles/how-to-use-local-https for a tutorial.


jcesarmobile:
There is a group called `MetadataGroup.DEPRECATED`, that I think you can use for this purpose without the need of changing texts manually. You just change the `MetadataGroup.EXPERIMENTAL` with `MetadataGroup.DEPRECATED` and it changes the text from `(experimental) Generates an SSL key & certificate` to `(deprecated) Generates an SSL key & certificate`, also the `experimental`/`deprecated` text has a different color than the rest of the description.

jcesarmobile (approved):

PR #5070, opened 2023-12-14
fix(cli): resolve vm2 security vulnerability

sean-perkins:
Ionic CLI currently reports a security vulnerability with the `vm2` dependency as a result of `superagent-proxy` → `proxy-agent`.

The [`superagent-proxy`](https://github.com/TooTallNate/superagent-proxy) package is a smaller wrapper around `superagent` and `proxy-agent` that is not maintained. 

This PR removes the dependency for `superagent-proxy`  and upgrades to `proxy-agent@6` which no longer has the `vm2` dependency requirement. The `proxy-agent` package uses ESM & as a result a lot of other usages had to be migrated once `esmoduleInterop` was enabled. The proxy behavior that `superagent-proxy` was responsible for has been added here: https://github.com/ionic-team/ionic-cli/blob/4332a24af67b98424ff19083a6f9e2e09a66c0a8/packages/%40ionic/cli/src/lib/utils/superagent-proxy.ts

Here is the error message without `esmoduleInterop`:

<Details>
<pre>
node_modules/get-uri/dist/http.d.ts:5:8 - error TS1192: Module '"http"' has no default export.

5 import http_ from 'http';
         ~~~~~

node_modules/get-uri/dist/http.d.ts:6:8 - error TS1192: Module '"https"' has no default export.

6 import https from 'https';
         ~~~~~

node_modules/proxy-agent/dist/index.d.ts:3:8 - error TS1259: Module '"/Users/sean/Documents/ionic/ionic-cli/packages/@ionic/cli/node_modules/proxy-agent/node_modules/lru-cache/index"' can only be default-imported using the 'esModuleInterop' flag

3 import LRUCache from 'lru-cache';
         ~~~~~~~~

  node_modules/proxy-agent/node_modules/lru-cache/index.d.ts:869:1
    869 export = LRUCache
        ~~~~~~~~~~~~~~~~~
    This module is declared with 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

Found 3 errors in 2 files.

Errors  Files
     2  node_modules/get-uri/dist/http.d.ts:5
     1  node_modules/proxy-agent/dist/index.d.ts:3
</pre>
</Details>


ltm (approved):

packages/@ionic/cli/ionic-cli-7.1.5.tgz
thetaPC, line 1: Remove this zip.

thetaPC (changes requested):

packages/@ionic/cli/.gitignore
```diff
@@ -3,3 +3,4 @@
 !jest.config.js
 !lint-staged.config.js
 *.d.ts
+*.tgz
```
thetaPC:
```suggestion
*.tgz

```
Missing new line

thetaPC (changes requested):

thetaPC (approved):
LGTM

PR #5065, opened 2023-11-17
fix(cordova): Adjust app paths for cordova-ios 7

jcesarmobile:
cordova-ios 7 changed the .ipa/.app paths
https://github.com/apache/cordova-ios/pull/1310

This PR adjust the path if executed in cordova-ios >= 7.

Created a new `getPackagePathCordova` function to not edit the existing `getPackagePath`, since it's exported maybe somebody is using it and could be breaking to change it. 
Only call the new method if the `getPackageJson` doesn't return an undefined `pkg`. 

closes https://github.com/ionic-team/ionic-cli/issues/5048

Issue #5048, opened 2023-10-16
iOS package path has changed, ionic run ios throws an error


webberig:
**Description:**
It appears that the latest version of XCode builds the app to a different folder (my version is 14.3.1). This causes an error when you run `ionic cordova run ios` because `native-run` is given the wrong path. 

The command executed and error:
```
> native-run ios --app platforms/ios/build/device/Zimmo.ipa --device
[native-run] ERR_UNKNOWN: Path 'platforms/ios/build/device/Zimmo.ipa' not found
[ERROR] An error occurred while running subprocess native-run.
``` 
However, my app is now in `platforms/ios/build/Debug-iphoneos` folder. This works:
```
native-run ios --app platforms/ios/build/Debug-iphoneos/Zimmo.ipa --device
```

**Steps to Reproduce:**
- In any Cordova + Ionic project, install the latest version of XCode.
- Run `ionic cordova run ios --device`

**Output:**
[native-run] ERR_UNKNOWN: Path 'platforms/ios/build/device/Zimmo.ipa' not found
[ERROR] An error occurred while running subprocess native-run.

**My `ionic info`:**
```
Ionic:

   Ionic CLI                     : 7.1.1 (/Users/mathieu/.nvm/versions/node/v18.12.1/lib/node_modules/@ionic/cli)
   Ionic Framework               : @ionic/angular 7.1.0
   @angular-devkit/build-angular : 16.2.1
   @angular-devkit/schematics    : 16.2.1
   @angular/cli                  : 16.2.1
   @ionic/angular-toolkit        : 7.0.0

Cordova:

   Cordova CLI       : 12.0.0 (cordova-lib@12.0.1)
   Cordova Platforms : android 12.0.1, ios 7.0.1
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 5.0.0, (and 21 other plugins)

Utility:

   cordova-res                          : not installed globally
   native-run (update available: 1.7.3) : 1.7.2

System:

   Android SDK Tools : 26.1.1 (/Users/mathieu/Library/Android/sdk)
   ios-deploy        : 1.12.1
   ios-sim           : 8.0.2
   NodeJS            : v18.12.1 (/Users/mathieu/.nvm/versions/node/v18.12.1/bin/node)
   npm               : 8.19.2
   OS                : macOS Unknown
   Xcode             : Xcode 14.3.1 Build version 14E300c
```

**Other Information:**
[...]

PR #5079, opened 2024-02-16
fix(cordova): Update build output path

MaximBelov:
Instead of build/emulator, we'll have build/Debug-iphonesimulator and instead of build/device, we'll have build/Release-iphoneos.

https://github.com/apache/cordova-ios/pull/1310


pinguluk:
@sean-perkins @liamdebeasi @mhartington @jcesarmobile can we please merge this fix already? it's been over a year since the issue appeared with the release of cordova ios 7.0.0

https://github.com/ionic-team/ionic-cli/issues/5048

EDIT:
it seems that there's already an open pull request and better, as it verifies the cordova version
https://github.com/ionic-team/ionic-cli/pull/5065

PR #5103, opened 2025-02-08
docs(run.ts): --livereload

rbalet:
## Description
`--livereload` into `--live-reload`


jcesarmobile:
Ionic's `--livereload` is different from Capacitor's `--live-reload`, it's a param for the server, it's not passed to Capacitor's command, so `--livereload` is the correct name.

PR #5104, opened 2025-03-03
fix(angular): change default project type to Standalone

brandyscarney:


packages/@ionic/cli/src/commands/start.ts
```diff
@@ -512,17 +512,17 @@ Use the ${input(
       const angularMode = await this.env.prompt({
         type: 'list',
         name: 'standalone',
-        message: 'Would you like to build your app with NgModules or Standalone Components? \n Standalone components are a new way to build with Angular that simplifies the way you build your app. \n To learn more, visit the Angular docs:\n https://angular.io/guide/standalone-components\n\n',
+        message: 'Would you like to build your app with Standalone Components or NgModules? \n Standalone components are the default way to build with Angular that simplifies the way you build your app. \n To learn more, visit the Angular docs:\n https://angular.dev/guide/components\n\n',
```
brandyscarney, line 515: I updated this guide because the old one is from v17, but if we would prefer linking to a NgModules guide there is one here: https://angular.dev/guide/ngmodules/overview#

jcesarmobile (approved):

ShaneK (approved):

PR #5128, opened 2026-06-16
feat(v8): foundation for the major 8.0 release

gnbm:
## Overview

Establishes the foundation for the **v8 major release** of `@ionic/cli`: the breaking-change documentation, the end-to-end major-release process, automated GitHub Releases, a hardened CI pipeline, full cross-platform support, and the TypeScript 5.9.3 upgrade.

## Breaking-change documentation

- Added `BREAKING.md` to log the v8 command removals, seeded with one entry per deprecated command (Appflow / Enterprise / Cordova), ready for each removal PR to fill in.
- Introduced an archive convention so future majors are filed cleanly.

## Major-release process

- Documented how to cut a major release end-to-end: the publish mechanism, how `@ionic/cli` reaches `8.0.0`, the path to publish from this branch, and how to validate safely before releasing (works on both Windows and macOS).

## GitHub Releases on publish

- Publishing now creates a GitHub Release automatically for every release, so the previously stale Releases page stays current.
- Achieved by upgrading the release tooling from **Lerna 3 → 5**.

## Safer, modernized CI

- Locked all GitHub Actions to specific verified versions for supply-chain security.
- Updated them to current releases, including replacing a cache action that GitHub had retired.

## Works on Windows and macOS

- Fixed pre-existing failures so the build, lint, and tests all pass on both Windows and macOS — previously lint and several tests only worked on Linux/macOS.

## TypeScript 5.9.3 upgrade

- Upgraded the whole codebase from TypeScript 4.8 → **5.9.3**, along with the test toolchain (jest / ts-jest to v29), with no loss of functionality and continued Node 16 support.

## Verification

✅ Certified green — build, lint, and the full test suite pass on **Node 16, 18, and 24**.

> [!IMPORTANT]
> **Squash-merge this PR** so the foundation work lands as a single commit on `major-8.0`.

### DIFF
--- a/.github/RELEASING.md
+++ b/.github/RELEASING.md
@@ -136,13 +136,26 @@ npm run test
 #    Drift guard: a green build of @ionic/discover (no publisher.ts "broadcast"
 #    TS2322) confirms the netmask@2.0.2 + @types/netmask@2.0.5 pins resolved.
 
-# 4. Safe release dry-run — no tag, no push, no publish
-npm run publish:testing
+# 4. Non-publishing version smoke test — exercises the Lerna 5 version path
+#    WITHOUT publishing to npm or touching git. `--allow-branch` overrides the
+#    lerna.json `allowBranch: stable` guard for this one local run.
+npx lerna version --conventional-commits --no-push --no-git-tag-version \
+  --allow-branch "$(git branch --show-current)" --yes
+git checkout -- .   # discard the version/CHANGELOG file writes lerna just made
 ```
 
-`publish:testing` is the safest smoke test — it exercises the Lerna 5 publish
-path without tagging or pushing. **Run it before the first live v8 cut** to
-confirm the 3 → 5 upgrade behaves (flag/config changes across Lerna 4 and 5).
+> **`npm run publish:testing` is NOT a dry run — it publishes to npm.**
+> `lerna publish` runs an `npm publish` (the `@ionic/*` packages under the
+> `testing` dist-tag); `--no-push`/`--no-git-tag-version` only suppress *git*
+> actions, not the registry publish. It also only runs from `stable` (the
+> `allowBranch` guard). Use step 4 above for a safe local check. Run
+> `publish:testing` only when you actually intend to push a `testing` prerelease
+> to npm, with valid npm credentials.
+
+The step-4 smoke test also reveals the **computed version bump** from the
+branch's conventional commits. Note it will show a `patch`/`minor` bump until
+the branch carries breaking-change commits (`feat!:` / `BREAKING CHANGE:`) —
+those, not `BREAKING.md`, are what make `@ionic/cli` land on `8.0.0` (see §1).
 
 > **Known platform note (pre-existing, unrelated to the release tooling):** on
 > **Windows**, one test in `integrations/cordova` asserts POSIX (`/`) path
--- a/.github/RELEASING.md
+++ b/.github/RELEASING.md
@@ -137,11 +137,11 @@ npm run test
 #    TS2322) confirms the netmask@2.0.2 + @types/netmask@2.0.5 pins resolved.
 
 # 4. Non-publishing version smoke test — exercises the Lerna 5 version path
-#    WITHOUT publishing to npm or touching git. `--allow-branch` overrides the
+#    WITHOUT publishing to npm or touching git. --allow-branch "*" overrides the
 #    lerna.json `allowBranch: stable` guard for this one local run.
-npx lerna version --conventional-commits --no-push --no-git-tag-version \
-  --allow-branch "$(git branch --show-current)" --yes
-git checkout -- .   # discard the version/CHANGELOG file writes lerna just made
+#    Single line + quoted "*" so it runs the same in PowerShell, cmd, bash, zsh.
+npx lerna version --conventional-commits --no-push --no-git-tag-version --allow-branch "*" --yes
+git checkout -- packages   # discard the version/CHANGELOG file writes lerna just made
 ```
 
 > **`npm run publish:testing` is NOT a dry run — it publishes to npm.**
--- a/.github/RELEASING.md
+++ b/.github/RELEASING.md
@@ -106,8 +106,8 @@ Requirements for `--create-release github` to work in CI:
 
 These steps are **cross-platform** — every command is `npm`/`npx`-based and runs
 identically in PowerShell, `cmd`, bash, and zsh (Windows and macOS). Use
-**Node 18** (the version CI runs — see `cd.yml`/`ci.yml`); newer Node majors can
-trip the bundled jest 26 worker on some platforms.
+**Node 18** to match CI (see `cd.yml`/`ci.yml`); the jest 29 toolchain this
+branch ships also runs green on Node 16 and 24.
 
 > **Order matters.** This is a `lerna bootstrap` monorepo: per-package
 > devDependencies (including `@types/node`) are installed and hoisted by
--- a/.github/RELEASING.md
+++ b/.github/RELEASING.md
@@ -17,10 +17,6 @@ Release is **conventional-commit driven and automatic**:
   `stable` branch**.
 - It runs `npm run publish:ci`, defined in [`package.json`](../package.json):
 
-  ```jsonc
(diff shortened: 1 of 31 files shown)


.github/RELEASING.md
```diff
@@ -0,0 +1,164 @@
...
+3. Follow the normal `develop` → `stable` promotion. The push to `stable`
+   triggers `cd.yml`, which computes the major bump, tags, publishes to npm, and
+   creates the GitHub Releases.
+
+### 2c. Optional: ship a v8 release candidate first
+
+Use a prerelease on a `next` dist-tag before GA so consumers can test:
+
+```bash
+# Dry, local prerelease (already defined, never pushes):
```
ShaneK:
```suggestion
# A real prerelease under the `testing` dist-tag (publishes to npm, no git push):
```

This contradicts the warning further down at line 147, which correctly says `publish:testing` is NOT a dry run. `lerna publish` runs `npm publish`, so this publishes the `@ionic/*` packages to npm under the `testing` tag. `--no-push` only suppresses the git push, not the registry publish. Looks like [672bea26](https://github.com/ionic-team/ionic-cli/commit/672bea26) fixed the line-147 wording but left this comment stale. I'd want this fixed before approving, since reading it as a safe local command could cause an accidental publish. Worth pointing readers at the step-4 smoke test for the actual safe check.

.github/RELEASING.md
```diff
@@ -0,0 +1,164 @@
...
+  step. [`cd.yml`](workflows/cd.yml) provides it as `${{ github.token }}` — the
+  built-in workflow token, which is sufficient because the workflow already
+  grants `contents: write`. **No dedicated secret is required.**
+
+## 4. Pre-publish validation (always, but especially after the Lerna upgrade)
+
+These steps are **cross-platform** — every command is `npm`/`npx`-based and runs
+identically in PowerShell, `cmd`, bash, and zsh (Windows and macOS). Use
+**Node 18** (the version CI runs — see `cd.yml`/`ci.yml`); newer Node majors can
+trip the bundled jest 26 worker on some platforms.
```
ShaneK: This PR upgrades jest to 29 everywhere (`^29.7.0` across all packages), so the "bundled jest 26 worker" reference is stale on arrival. Worth updating to jest 29, and I think the caveat itself may not hold anymore since jest 29 supports newer Node than 26 did. Worth double-checking before keeping it.

packages/@ionic/discover/package.json
```diff
@@ -21,29 +21,29 @@
...
     "build": "npm run clean && tsc",
     "watch": "tsc -w --preserveWatchOutput",
     "test": "jest --maxWorkers=4",
     "prepublishOnly": "npm run build"
   },
   "dependencies": {
     "debug": "^4.0.0",
-    "netmask": "^2.0.2",
-    "tslib": "^2.0.1",
+    "netmask": "2.0.2",
```
ShaneK: There's no committed lockfile in this repo, so `^2.0.2` was already resolving to the newest safe 2.x (currently `2.1.1`). Pinning to exact `2.0.2` freezes it at the floor and blocks future 2.x patches, including security ones, which is the opposite of what we want. I see the commit body pins it to dodge the `TS2322` type break from netmask `2.1.x`. If that's the driver, could we pin `@types/netmask` instead (the actual source of the type break) and keep the caret on the runtime `netmask`?

BREAKING.md
```diff
@@ -0,0 +1,128 @@
...
+workflows. The Appflow, Enterprise, and Cordova commands have been removed.
+
+- [Removed Commands](#removed-commands)
+
+### Removed Commands
+
+The following commands have been removed. Each entry records **what was
+removed**, **why**, and the **migration path / alternative**.
+
+> **Note:** The removal PRs are authored and coordinated separately. The entries
```
ShaneK, line 26: These `<TODO:>` placeholders read as intentional seeds given the doc explains it right here and the removals land in separate PRs, so no objection to the approach. One thought: a CI guard that fails if `BREAKING.md` still contains `<TODO` on the `stable` branch would make the "each removal PR fills it in" contract enforceable, rather than risking `<TODO: alternative / link.>` shipping to users at `8.0.0`. Just something to think about, no action required.

ShaneK (changes requested):
Solid foundation start, but some requests

.github/RELEASING.md
gnbm: @ShaneK  Done!

packages/@ionic/discover/package.json
gnbm: @ShaneK  Done!

BREAKING.md
gnbm: @ShaneK, I tried adding a different approach to this, so please let me know your thoughts

ShaneK (approved):
Thanks for resolving those issues! Looks like a good foundation

.github/RELEASING.md
```diff
@@ -0,0 +1,164 @@
...
+Release is **conventional-commit driven and automatic**:
+
+- [`.github/workflows/cd.yml`](workflows/cd.yml) runs **only on push to the
+  `stable` branch**.
+- It runs `npm run publish:ci`, defined in [`package.json`](../package.json):
+
+  ```jsonc
+  "publish:ci": "lerna version -m 'chore(release): publish [skip ci]' --exact --conventional-commits --yes --create-release github && lerna exec --no-private --since HEAD~ -- npm publish --provenance"
+  ```
+
```
thetaPC: I would recommend leaving this out, if the command gets updated then the dev would have to remember to update it here. You're already pointing to the `package.json` so that should be enough.


```suggestion
```

.github/RELEASING.md
```diff
@@ -0,0 +1,164 @@
...
+  - `lerna version --conventional-commits` computes each package's next version
+    **from commit messages**, tags, and pushes the release commit.
+  - `--create-release github` makes Lerna create the **GitHub Release** for each
+    tag (see **3** — requires Lerna 5 and a `GH_TOKEN`).
+  - `lerna exec ... npm publish --provenance` publishes the changed packages to
+    npm.
+- `lerna.json` pins `"allowBranch": "stable"`, so `lerna version` refuses to run
+  from any other branch.
+- After publishing, `cd.yml` builds and pushes a Docker image to GHCR.
```
thetaPC: What's GHCR? It can be beneficial to write out what it stands for.

.github/RELEASING.md
```diff
@@ -0,0 +1,164 @@
...
+  one that reaches `8.0.0`. Utility `@ionic/*` packages bump independently based
+  on their own commits.
+
+### 2b. Path to publish from `major-8.0`
+
+`cd.yml` only publishes from `stable`. The `major-8.0` work branch is **not**
+auto-published. Recommended path:
+
+1. Land all removal PRs (with breaking-change commits) onto `major-8.0`.
+2. Merge / rebase `major-8.0` → `develop`.
```
thetaPC: Is this a normal merge? It can help distinguish by adding it as some repos use squash merge.


```suggestion
2. Normal merge / rebase `major-8.0` → `develop`.
```

.github/RELEASING.md
```diff
@@ -0,0 +1,164 @@
...
+- A **`GH_TOKEN`** must be present in the environment of the `lerna version`
+  step. [`cd.yml`](workflows/cd.yml) provides it as `${{ github.token }}` — the
+  built-in workflow token, which is sufficient because the workflow already
+  grants `contents: write`. **No dedicated secret is required.**
+
+## 4. Pre-publish validation (always, but especially after the Lerna upgrade)
+
+These steps are **cross-platform** — every command is `npm`/`npx`-based and runs
+identically in PowerShell, `cmd`, bash, and zsh (Windows and macOS). Use
+**Node 18** to match CI (see `cd.yml`/`ci.yml`); the jest 29 toolchain this
```
thetaPC: Since node 18 is needed, can we make it required through the `package.json`?

.github/RELEASING.md
gnbm: @thetaPC GHCR is a standard and stands for GitHub Container Registry

thetaPC (approved):
LGTM

